PR #15, opened 2016-04-20
fixed storageManager config error

walkskyer:


artf:
thanks @walkskyer 
You're right! The `remoteStorage` part is wrong and your fix is right for the `remote` configuration, but actually I would like to keep `local` as default storage in demo. So if you would restore `storageType` back to `local` and remove `storageManager` part (you don't need it with local) I'll merge it :)

artf:
thanks @walkskyer

walkskyer:
@artf Thank you! It's more helpful for me!

PR #16, opened 2016-05-04
change the config path

walkskyer:


artf:
@walkskyer nice, thanks

Issue #66, opened 2017-04-08
How to load external css file on editor


narraressan:
Hi. Thank you for solving my problems on the editor this past few days,

But while I am finalizing my simple work, I decided to use skeleton css, but I can't figure out how to load css file on the editor? (inside the iframe) is it possible?

narraressan:
Well, I think I found a work around, I am not sure if it works 100% 

I added <link rel="stylesheet" href="css/bhs-skeleton.css"> to editor using setComponent...

allmonday:
Maybe you can append it in head of iframe.
```js
var iframe = editor.Canvas.getFrameEl();
iframe.contentDocument.head.appendChild(/* your link tag here */);
```

artf:
Actually, [setComponents](https://github.com/artf/grapesjs/wiki/API-Editor#setcomponents) is the way to go. The problem is that it overwrites the whole template. I think I'll add `addComponents` on the next release but it'll be basically the same as:
```js
editor.getComponents().add('<link rel="stylesheet" href="...">');
```
@allmonday your snippet actually will add the link tag but the problem it won't update the template state, so when you'll call `editor.getHtml()` you won't see it inside the code.

prtjack:
hey! @artf 
How can i save an existing template with its own external CSS properties? I tried the both options: 
1. By placing entire HTML and CSS under the <div id="gjs"></div>.
2. By using editor.setComponents() i can achieve my purpose but its very complicated to minify the whole CSS and place under editor.setStyle().

So what is the best option to export  an existing template with it's CSS.

artf:
Hi @prtjack can you elaborate why the 1st option is not working? Can you post some code?

prtjack:
Hey! @artf 
Sorry for this  extremely late reply. I wanted to load my existing templates dynamically. So, what is the best workaround for me?

likhamahudda:
@octavebenil 
Thanks very much its working fine.
canvas: {
scripts: ['https://ajax.googleapis.com/ajax/libs/jquery/3.1.1/jquery.min.js'],
styles: ['https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css']
}

gitwithsarthak:
> <img alt="@likhamahudda" width="40" height="40" src="https://avatars.githubusercontent.com/u/23294648?s=88&amp;u=ee4fb6cff446fcdfa333724cc5346e6586e44970&amp;v=4">

[...]

PR #67, opened 2017-04-11
enable scripts in iframe head

allmonday:
when i set allowscript option to 1, I tried to add required js library at head using:
```javascript
// ...
editor.render();
var f = editor.Canvas.getFrameEl();
scripts.map((script)=>{
    f.contentDocument.head.appendChild(script);
});
```
however it will fail cause the render body is executed before the code above.
I hope to have a config option to insert the scripts I need.

with the PR, now it can be configured in grapesjs.init 

```javascript
var editor  = grapesjs.init({
    // other options
    canvas: {
        scripts: [
            "https://cdnjs.cloudflare.com/ajax/libs/angular.js/1.6.1/angular.min.js"
        ]
    },
}
```

### DIFF
--- a/src/canvas/config/config.js
+++ b/src/canvas/config/config.js
@@ -6,5 +6,8 @@ define(function () {
 		// Coming soon
 		rulers: false,
 
+		// append scripts in head
+		scripts: []
+
 	};
 });
--- a/src/canvas/config/config.js
+++ b/src/canvas/config/config.js
@@ -6,7 +6,10 @@ define(function () {
 		// Coming soon
 		rulers: false,
 
-		// append scripts in head
+		/*
+		 * append scripts in head of iframe before renderBody content
+		 * need to manually maintain the same scripts in cms's render template
+		*/ 
 		scripts: []
 
 	};
--- a/src/canvas/view/CanvasView.js
+++ b/src/canvas/view/CanvasView.js
@@ -31,6 +31,31 @@ function(Backbone, FrameView) {
     },
 
     /**
+     * Insert scripts into head, it will call renderBody after all scripts loaded or failed
+     * @private
+     */
+    renderScripts: function () {
+        var frame = this.frame;
+        var that = this;
+
+        frame.el.onload = function () {
+          var scripts = that.config.scripts,
+              counter = 0;
+
+          scripts.map(function (scriptUrl) {
+            var script = document.createElement('script');
+            script.type = 'text/javascript';
+            script.src = scriptUrl;
+            script.onerror = script.onload = function () {
+              counter++;
+              if (counter == scripts.length) that.renderBody();
+            };
+            frame.el.contentDocument.head.appendChild(script);
+          });
+        };
+    },
+
+    /**
      * Render inside frame's body
      * @private
      */
@@ -198,7 +223,11 @@ function(Backbone, FrameView) {
         this.model.get('frame').set('wrapper', this.wrapper);
         this.$el.append(this.frame.render().el);
         var frame = this.frame;
-        frame.el.onload = this.renderBody;
+        if (this.config.scripts.length === 0) {
+          frame.el.onload = this.renderBody;
+        } else {
+          this.renderScripts(); // will call renderBody later
+        }
       }
       var ppfx = this.ppfx;
       toolsEl = $('<div>', { id: ppfx + 'tools' }).get(0);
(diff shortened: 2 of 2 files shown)


artf:
Hi @allmonday and thanks for your efforts, unfortunately, you've not considered the state of the template and it's almost the same as I explained you here #66 

There is also a problem with async loading to take care, for instance:
```js
...
scripts: [
  ".../jquery.js",
  ".../jquery-tiny-plugin.js" // this probably will be loaded first and so you'll get '$' ref error
        ]
```
Can you past here your template (at least an example) so I can try to give you the workaround?

Have you tried with:
```js
var dc = editor.getComponents();
dc.add('<script src="..."></script>');
editor.render();
```

allmonday:
hi @artf , I've tried `editor.getComponents();`, but has a problem: 

first version, set `autorender: 0` and add new script tag, then call render(), It does not work.
```javascript
// app.js
var editor = grapesjs.init({
    allowScripts: 1,
    autorender: 0,
    container : '#gjs',
    components: `<div class="txt-red">Hello__ world!</div>`,
    style: '.txt-red{color: red;}',
});

var dc = editor.getComponents();
dc.add('<script src="//cdn.bootcss.com/jquery/2.2.4/jquery.min.js"></script>');
editor.render();
```
second version, It works and insert the script.
```javascript
var editor = grapesjs.init({
    allowScripts: 1,
    // autorender: 0,
    container : '#gjs',
    components: `<div class="txt-red">Hello__ world!</div>`,
    style: '.txt-red{color: red;}',
});

var dc = editor.getComponents();
dc.add('<script src="//cdn.bootcss.com/jquery/2.2.4/jquery.min.js"></script>');
// editor.render();
```
but here comes a problem, if I edit the page, after save, then reload the page (reading it from localstorage),  the jquery script will be inserted twice.

first time:  (just ignore the first jquery it's on the page itself, not iframe)
image

edit something and reload:
image

allmonday:
I understand that insert a list of script will cause the async loading which will cause dependency error, and I'll explain why I make these changes:

I need to add some customrized blocks which  has snippets of script,
(and make it a reference like component, which means if I change the template and script, all the page using it will be changed)
eg:

```javascript
// blockManager.blocks:
    {
            id: 'my-first-block',
            label: 'leadform',
            attributes: {class:'fa fa-link'},
            content: `
            <div class="row">
                <div id="form-demo" ng-cloak class="my-block">
                    <leadform></leadform>   <!-- wrap it as a element tag -->
                </div>
            </div>
            <script>
                var s = document.createElement("script");
                s.type = "text/javascript";
                s.src = "/javascripts/lead.js";  // read it from server
                document.body.appendChild(s);
            </script>
}
```

 ```javascript
// lead.js
angular.module("app", [])
    .component("leadform", {
        templateUrl: "/api/leadtemplate/",    // read it from server 
        controller: function LeadFormController($http, $scope) {
        // .. some code here
    });
angular.bootstrap(document.getElementById("form-demo"), ["app"]);
```

then I can manage all the blocks.

In production, I can prepare all the library at head of template and insert `gjs-html` content (and css ..) into it, then render.

but in grapesjs editor, inserting necessary library is not that easy. using `getComponent`, `add` is not that fit to it because of the order.

That's why I want to insert scripts into iframe's head, I wanna a clear order: loading necessary scripts, then renderBody. (create iframe -> insert scripts -> renderBody)

btw, the defect is also obvious, user need to add the scripts at their own template (but maintaining script and css tag in gjs-html's edit and save cycle is kind of complicated).

**Comment by artf** [comment]
Thanks @allmonday for well explaining your case, which is pretty interesting.
Anyway, if you add templates with `components:` option, I think this should already work for you:
```js
var editor = grapesjs.init({
  ...  
  components: `<div class="txt-red">Hello__ world!</div>
   <script src="..."></script>`,
  ...
});
```
About twice inserted thing, is happening because you store the template, which already contains your inserted `script`,  then refresh (load) and add the `script` again. So, in this case, you should add only with not yet saved templates.

Overall, I think the idea of injecting scripts independently from the template (as you'll manage them outside manually) could be a nice feature, but there is still a problem with async loading to fix (you should be able to do it by appending next scripts inside each `onload` callback)

allmonday:
yes, the vendor scripts may be required by other blocks( in my design ), I used to try to check whether the vendor is existed in each block's script, if not then append it, but this makes the block keeping too much business unrelated code.

artf:
Thanks @allmonday 👍

PR #76, opened 2017-04-24
add content-type: application/json; charset=utf-8 for storing data

allmonday:
hi @artf, 3Q for you great work, this project helps me a lot.
In my use I find the default store parameter of remote storage has some problem with asp.net webapi.

Backend can automatically handle a json body, turn it into model then handle it, the current content-type is x-www-form-urlencoded, body string is like form post which require some extra operation for backend guys.  

```c#
// orm attributes
[jsonProperty("gjs-css")]
public string Css { get; set; }
```
That's why I add a new parameter, to be able to send a json to the backend.


artf:
Thanks @allmonday seems totally legit to me, but what about:
```js
contentType: 'content-type-string',
```
will be `x-www-form-urlencoded` in case is not defined

allmonday:
Yes, I've considered this solution before, I wanna it be easy to use cause the content type is also related to `data` (stringify or not) , there will be another parameter about how to deal with the `data`
It may looks less flexible though, but much easy to use and meet the major requirement  :)

PR #80, opened 2017-04-27
fix dom_components script crash

vladflorescu94:
The application is crashing if you create a component where script gets a function.


artf:
Good catch, thanks @vladflorescu94

PR #105, opened 2017-06-08
utils module case mismatch

givanz:
Changed utils from capitalize to lowercase to match directory name format to avoid the following error

ERROR in ./src/editor/model/Editor.js
Module not found: Error: Can't resolve 'Utils' in 'grapejs-git/grapesjs/src/editor/model'


artf:
thanks @givanz

PR #202, opened 2017-08-03
Temporarily adding components via the Sorter suppresses the change event and includes a flag that can be checked in other event handlers

### LINKED ISSUE #189: Add Component event is fired twice when dragging blocks onto the canvas
I'm trying to invoke some custom code when a component is added to the canvas. I was hoping to use the `component:add` event as this seems to cater for manually loading content (via `editor.setComponents`), as well as dragging blocks onto the canvas from the block manager. However I've noticed that the event is firing twice when dragging and dropping - it seems as though it fires once as soon as you mouse over the canvas, and then it fires again when you drop.

I've got some code below which can be used to replicate the issue. I'm using version 0.9.8.

```` html
<!DOCTYPE html>
<html>
<head>
  <title>GrapesJS Test</title>
  <link rel="stylesheet" href="/scripts/grapesjs/css/grapes.min.css" />
  <script src="/scripts/jquery-2.1.4.min.js"></script>
  <script src="/scripts/grapesjs/grapes.min.js"></script>
</head>
<body>
  <div>
    <div id="gjs"></div>

    <script type="text/javascript">
      var editor = grapesjs.init({
        container: '#gjs',
        autorender: false,
        fromElement: false,
        storageManager: {
          autosave: false,
          autoload: false
        }
      });

      var comps = editor.DomComponents;
      var defaultType = comps.getType('default');
      var defaultModel = defaultType.model;
      var defaultView = defaultType.view;

      comps.addType('widget',
        {
          model: defaultModel.extend({
              defaults: Object.assign({},
                defaultModel.prototype.defaults,
                {
                  type: 'widget',
                  tagName: 'div',
                  content: '<img src="https://placehold.it/600x400?text=Widget+Placeholder"/>',
                  draggable: ['div'],
                  droppable: false,
                  editable: false,
                  attributes: {
                    widget: 'true'
                  }
                })
            },
            {
              isComponent: function(el)
              {
                if ((el.tagName == 'div' && el.getAttribute('widget') === 'true'))
                {
                  return { type: 'widget' };
                }

                return false;
              }
            }),

          view: defaultType.view
        });

      var blockManager = editor.BlockManager;

      blockManager.add('test-block',
        {
          label: 'Test block',
          content: {
            type: 'widget'
          }
        });

      editor.on('component:add', function(model) {
          console.log('add component');
          console.log(model.get('type'));
          console.log(model);
        });

      editor.render();

    </script>

  </div>
</body>
</html>
````

When I load the page and drag my block onto the canvas, I get the following console output (running Chrome):

```
add component
widget
n {cid: "c571", .... }
add component
widget
n {cid: "c583", ... }
```
Based on that cid property it looks like it's creating two copies of the component.

### PR DESCRIPTION
**Pull request by thecodefish**
Fixes #189.

This adds a 'temporary' flag to the event options when the component is temporarily added to the collection (and then immediately removed). These options are also propagated along the event chain properly.

The consequences of this is that the `changesCount` value is not updated, and autosave is not triggered. Plus, in any custom event handlers such as for `component:add`, you can now inspect the options parameter and take appropriate action. For example:

``` js
editor.on('component:add',
  function(model) {
    if (model.opt.temporary)  {
      return;
    }

    //your normal event handler code here
});
```

Specifically (for my own needs), this means that when you drag a Block onto the canvas, `change:changesCount` is only fired once when you actually drop the block, and `component:add` is still fired twice, however you can ignore the first trigger by checking the temporary option.

I decided against using `{silent:true}` as this seems to suppress *all* events which would probably have unintended consequences.

@artf Note that I haven't run any tests against this change as I can't get it working for me (Windows 10). Even after installing npm-run I get path problems and errors trying to execute the tests/build scripts (even before my changes).


artf:
thanks @thecodefish really appreciate your contribution, I've tested your branch and it works as expected. About tests in Windows, I think it's due to path variables, therefore I just added `cross-env` for the next release (hope to push soon)

adamlc:
I've just come across this issue and this fix works, probably worth mentioning in the docs?

artf:
@adamlc in the next release `component:add` are not triggered with `temporary` adds :)

dmac1025:
I'm wondering if there should be a check for the temporary flag in the components.js onAdd event like below:
image

Or if I am missing something, because it seems with the latest release the CSS class is generated twice for a custom component when added to the canvas.  Great project by the way.
Thanks.

artf:
Thanks @dmac1025 but if you encountered a new bug please create a new issue and follow its template

PR #211, opened 2017-08-05
Adding remove function to Blocks manager

### LINKED ISSUE #194: Block API remove function
I would assume that the `remove()` function would be required in such project, there's adding, getting, but not removing selected block, I would like to remove button block, and add my own - so I can make my own styles for it.

uniacid:
@artf I added this on my end for a project utilizing grapesjs, this should presumably resolve #194


artf:
thanks @uniacid 👍

PR #213, opened 2017-08-05
Updating PropertyFileView file upload modal

### LINKED ISSUE #212: Style manager Property File View does not properly utilize open-assets command
I noticed while trying to override the built-in Asset manager uploader it was only working on image components/tags itself and anything else that used the Asset manager uploader for example an background-image (Style Manager property) while it would open the Asset manager modal it was not doing so in the same manner as the Image component would and thus didn't allow me to override the built-in uploader.

uniacid:
@artf This is to resolve #212 an update to the openAssetManager function in PropertyFileView to utilize the open-assets command itself to open the modal.

Reason for this being as I was overriding the built-in image upload function and modal with my own but I noticed when using the same file uploader for a background image it wouldn't even hit my custom overloading uploader as it relies on it running the command function.

Let me know your thoughts on that issue and otherwise if this looks like a good fix to merge in.

### DIFF
--- a/src/style_manager/view/PropertyFileView.js
+++ b/src/style_manager/view/PropertyFileView.js
@@ -120,15 +120,20 @@ module.exports = PropertyView.extend({
    * */
   openAssetManager(e) {
     var that  = this;
-    if(this.modal && this.am){
+    var em = this.em;
+    var editor = em ? em.get('Editor') : '';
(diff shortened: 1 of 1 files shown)


artf:
Awesome, thanks @uniacid 👍

PR #261, opened 2017-08-29
Update index.js

Fi1osof:
Error: Cannot find module 'editor/model/editor'


artf:
Thanks @Fi1osof got also the notice from Travis 😁

PR #281, opened 2017-09-06
Sass variables

A----:
Moves variables into its own file, allows overriding by suffixing `!default`.

This has the following advantages:
1. apps that have SASS support can integrate easily Grapes by overridden some/all variables;
2. if an app uses Webpack with sass-loader, there's a catch with `url()` ( https://github.com/webpack-contrib/sass-loader#problems-with-url ), font path variable can now be overridden.

It's be to used as follows:
```scss
// 1. theming
$mainColor: $my-bg-color;
$fontColor: $my-fg-color;
$fontColorActive: $my-active-color;
// etc.

// ~package means "webpack path resolution of package", node_modules basically
$fontPath: '~grapesjs/src/styles/fonts';  // 2. font path
@import '~grapesjs/src/styles/scss/main';
```


artf:
Amazing, thanks 👍

A----:
What overriding four variables can do:
image

Geczy:
For anyone else that doesn't want to go through an hour of debugging, here's how to override variables in grapesjs:

```scss
$fontPath: '/~grapesjs/src/styles/fonts';
$fa-font-path: '/~grapesjs/src/styles/fonts';

@import '/~grapesjs/src/styles/scss/main';
```

Trick was to include fa-font-path aswell

PR #282, opened 2017-09-06
Pass element in config

A----:
Allow user to pass an HTML node instead of a selector for the container configuration.

Frameworks like Angular asks of their components to be reusable. Basically, you would to this instead of relying on a selector:

# Template
```html
<div #grapejs></div>
```

# Component
```typescript
@Component({
  […]
})
export class MyGrapeJSComponent implements AfterViewInit {

  @ViewChild('grapesjs') grapesjs;
  private editor: any;

  ngAfterViewInit() {
    this.editor = grapesjs.init({
      container: this.grapesjs.nativeElement // Reference to the DOM node
    });
  }

}
```


A----:
I did not see other places where `container` is used except one time where `el` was checked first, but that means that you can't rely for `container` to be a string anymore.

suchithmahadi:
Artur Arseniev
Need you help in a Custom Component&JS, Please help me out

artf:
thanks @A---- absolutely a must-have option, don't know why I haven't done this before :D

suchithmahadi:
Can you guide me to do a custom component block for Google Chart ?

artf:
wtf @suchithmahadi stop posting here, your problem it's not related with this PR...

A----:
Thanks for the quick merge!

artf:
@A---- no problem thanks to you, by the way, don't run `npm run build` for PRs as you get useless conflicts with minified files 😬

A----:
I thought so too, I followed https://github.com/artf/grapesjs#development
Also I needed a compiled version for tests.

Maybe remove the `v:patch` from the `build` goal and change the readme?

artf:
> Also I needed a compiled version for tests

You can run `npm test` for tests

The problem is not with `v:patch` but with the building itself, you may need to build because you need your custom minified GrapesJS but for PRs I think it's just not necessary. Maybe I should just add `Don't build for PRs`?!?  🤔

PR #288, opened 2017-09-07
don't show rte toolbar for elements with editable parent

vladflorescu94:
Avoid displaying multiple rte toolbars when double clicking an inner editable element


artf:
Hi @vladflorescu94 thanks for the PR but can I ask you to provide an example code (or template) with which I can see the problem, thanks

artf:
ping @vladflorescu94

PR #301, opened 2017-09-13
Use data URL to embed images in documents

A----:
This PR allows images to be embedded in the document using data URLs.
I spent nearly an hour, but I could not manage to write proper unit tests for this.

Also:
- Adds an undocumented `disabledUpload` options. There was a unit test for it and it worked, but coincidental.
- Do not require for `onUploadResponse` to be a String, can simply be an object.

Also, there's an undocumented `name` attribute that's understood by `AssertView`.


artf:
Amazing feature, thanks mate

Issue #355, opened 2017-09-30
Custom plugins does not overrides default commands


arthuralmeidap:
@artf ,
  Until versions 0.9.x , GrapesJS loads all it's internal's modules before  all the plugins are initialized.
  Since from version 0.10.2 you added a new module initialization after the plugins initialization. 
  You can see it here: https://github.com/artf/grapesjs/blob/v0.10.2/src/grapesjs/index.js#L73
  This behaviour has some side-effects, for example on grapesjs-preset-newsletter plugin because it tries to override the command 'export-template' but, when you initialized the modules after the plugins, this command is overridden  again with the default command.

arthuralmeidap:
I've created a PR with a fix to this behavior.
Take a look here: https://github.com/artf/grapesjs/pull/356

PR #356, opened 2017-09-30
Fix custom plugin default command overriding

arthuralmeidap:
@artf ,
  Here is my fix. Take a look and tell me what you think.

### DIFF
--- a/src/commands/index.js
+++ b/src/commands/index.js
@@ -196,15 +196,9 @@ module.exports = () => {
       if(c.em)
         c.model = c.em.get('Canvas');
 
-      return this;
-    },
+      this.loadDefaultCommands()
 
-    /**
-     * On load callback
-     * @private
-     */
-    onLoad() {
-    	this.loadDefaultCommands();
+      return this;
     },
 
     /**
@@ -258,7 +252,7 @@ module.exports = () => {
      * */
     loadDefaultCommands() {
       for (var id in defaultCommands) {
-        this.add(id, defaultCommands[id]);
+          this.add(id, defaultCommands[id]);
       }
 
       return this;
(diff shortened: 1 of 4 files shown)


arthuralmeidap:
I forgot to mention but this PR is related for the issue #355

artf:
Hi Arthur and thanks for the contribution, by the way, I can't accept it as it's important that `onLoad`  triggers after plugins and the reason is just above in the comment
```js
// Execute `onLoad` on modules once all plugins are initialized.
// A plugin might have extended/added some custom type so this
// is a good point to load stuff like components, css rules, etc.
```
If you want to fix 'default command overriding' I'd suggest you try to move `onLoad()` at the end of  `init()` (in https://github.com/artf/grapesjs/blob/dev/src/commands/index.js)

arthuralmeidap:
Hi @artf,
  Don't worry, I saw the comment but when I ran the tests and everything went fine, I give a try for this PR.
  Unfortunately, your suggestion won't work either because the modules are being loaded after the plugins.
  One question: Why can't the Command Module load the default commands on init? I have a solution using this approach but one of the tests says : "Main Commands - No default commands on init" and it fails.. 
  Looking forward to hear from you

artf:
> Why can't the Command Module load the default commands on init?

This is exactly what I meant with `If you want to fix 'default command overriding' I'd suggest you try to move onLoad() at the end of init()` 😄 
```js
init() {
    ...
    this.loadDefaultCommands(); // this is all the content from onLoad
    return this;   
}

// Remove onLoad
onLoad() {
    ...
```

arthuralmeidap:
Hi @artf ,
  I did what you suggested. Look and tell what you think about it.
  All commands are loaded at the end of Command init  and removed toLoad property to not override again the default commands.

artf:
Great 👍 , can you just bring `editor.getModel().loadOnStart();` under the plugin loader then I can merge

arthuralmeidap:
Ops, sorry! Fixed now!

artf:
Thanks Arthur ❤️

Issue #369, opened 2017-10-03
Usage from source failed


ateshuseyin:
I tried to use grapesjs from source by using `require('grapesjs/src')` to debug both my source and grapesjs. But could not success because of below issues;

- There is some requires like `require('plugin_manager')`. These can be passed by webpack config but it is not a nice solution.
- '<# VERSION #>' preprocessing should be done by git hooks, not compile time.
- There is no guide about usage from source.

I think this issue is very imported to identify the bugs of grapesjs.

ateshuseyin:
At DevicesView.js:5 there is underscore usage w/o require: `template: _.template(``

ateshuseyin:
I just notice that you are using ProvidePlugin. That is a solution but these kind of configurative solutions are a limitation to grapesjs users which are using from source.

artf:
correct @ateshuseyin I just never configured it to work properly from source, indeed have to move few configurations from webpack to npm. Appreciate your help 👍

PR #370, opened 2017-10-03
webpack resolve modules src removed

ateshuseyin:
In this PR, I removed local requires as modules which I described before in #369. 

related to #369


artf:
Thanks @ateshuseyin 
I'm not a big fan of `../../` are ugly and don't even scale well, maybe we can use npm local modules?!?
eg. 
```js
“dependencies”: {
     ...
   “gjs”: “file:src”,
 }
```

artf:
any update @ateshuseyin?

ateshuseyin:
I tried local modules with webpack but it cannot resolve local modules without defining resolve statement in webpack config.

You are right `../..` are ugly but it is just the solution.

artf:
well, I think it's possible to create an alias for the `gjs` module in webpack config :)

artf:
By reading this https://gist.github.com/branneman/8048520 seems like there are no solid solutions for this issue...

tommedema:
@artf I realize this is from 2017, but the `var Article = require.main.require('app/models/article');` solution seems pretty solid to me.

artf:
@tommedema webpack supports `require.main` but not `require.main.require`

Piterden:
Why not use alias like the next inside a webpack's config?
```js
const path = require('path')

function resolve (dir) {
  return path.join(__dirname, '..', dir)
}

module.exports = {
  resolve: {
    alias: {
      '@': resolve('src'),
    },
  },
}
```
then import it like:
```js
import ItemView from '@/navigator/view/ItemView.js'
```

artf:
@Piterden thanks for the suggestion but how exactly would you implement it with npm as there is no concept of aliases?

Piterden:
https://www.npmjs.com/package/module-alias

artf:
Well cool, but that warning basically discourages the main point of the PR, using grapesjs from source (usually inside other npm modules) :(

Piterden:
Nude node is so uncomfortable )))

PS I don't know why, but I use only `index.js` files under `src` for my last projects...

PR #414, opened 2017-10-12
Trigger events when editing a component with the rich text editor.

tncalvert:
This adds two new events: 'rte:attach' and 'rte:detach'. They are triggered when a user starts or stops editing the text of a component using the rich text editor.

The events are triggered on the Editor model and listeners receive the model that is being edited.


artf:
Hey Tim thanks for this PR, but I've just pushed the all-new RTE module 😂  but don't worry, it includes basically your same events, 'rte:enable' and 'rte:disable' (I pass the view instead of the model as the argument)

tncalvert:
No problem. Thanks for all your work; grapesjs is excellent.

Issue #214, opened 2017-08-06
GrapesJS does not work in IE11


altalker:
GrapesJS does not work in IE11 as many of the JS function are not supported (e.g. remove().. , "`", blob.map..

artf:
Can you show me exactly what kind of errors do you have?

altalker:
1. ![image](https://user-images.githubusercontent.com/29933877/29083361-5b1c84f4-7c1d-11e7-82d6-a653d2a8fc1e.png)
  content: `<di

2. 
![image](https://user-images.githubusercontent.com/29933877/29083505-d47a6dac-7c1d-11e7-8c83-84eaec9ed0fa.png)
3. 
![image](https://user-images.githubusercontent.com/29933877/29083531-e8fb12fe-7c1d-11e7-937f-24fdbcfbb599.png)

altalker:
This is when I use IE 11 when items are dragged into the Canvas or using filestack.  MS Edge it is ok. (JavaScript ECMA compatibility differences)

artf:
thanks @altalker unfortunately, I'm not able to check those issues right now but hope to do so soon

nicksav:
Actually have dnd broken on Edge as well.
Just doesn't work

artf:
hey @nicksav can you post please your errors

umangpasricha:
I am facing the same problem when using GrapesJs on IE11. Even [http://grapesjs.com/demo.html](URL) is not working fine. I am getting these errors:

![image](https://user-images.githubusercontent.com/25193657/30860372-88441472-a2e4-11e7-9b79-1f0d4cec2355.png)
when I drag the components/blocks to the canvas(even before I drop them).

artf:
I think it's clear that we have to add some polyfill for those methods, but at the moment, it's just not my priority (can't even test them properly), so... PRs are welcome

arthuralmeidap:
@altalker wich Windows version are you using?

umangpasricha:
I added polyfill for these methods and now it is working fine on IE too. 
Here is the code I added for 'assign' method in the .js file I am using:

**if (typeof Object.assign != 'function') {
		  Object.defineProperty(Object, "assign", {
		    value: function assign(target, varArgs) { 
		      'use strict';
		      if (target == null) { 
		        throw new TypeError('Cannot convert undefined or null to object');
		      }
		      var to = Object(target);
		      for (var index = 1; index < arguments.length; index++) {
[...]

PR #435, opened 2017-10-19
switched to babel-preset-env

ateshuseyin:
In this PR, I use babel-preset-env to provide best browser compatibility. I am mac user so I cannot test with IE but it should work. Can any one test this with IE.
relates #214


artf:
Awesome 👍

PR #437, opened 2017-10-19
Specify GET Method to Load templates

santiph:
RemoteStorage is using `POST` to load templates
Issue: https://github.com/artf/grapesjs/issues/436

### DIFF
--- a/src/storage_manager/model/RemoteStorage.js
+++ b/src/storage_manager/model/RemoteStorage.js
@@ -74,15 +74,7 @@ module.exports = require('backbone').Model.extend({
   },
 
   load(keys, clb) {
-    let queryString = '';
-    keys.forEach(function(key, index, keys) {
-      queryString += key;
-      if (index < keys.length - 1) {
-        queryString += ',';
-      }
-    }, this);
-
-    this.request(`${this.get('urlLoad')}?keys=${queryString}`, {method: 'get'}, clb);
+    this.request(this.get('urlLoad'), {method: 'get'}, clb);
   },
 
   /**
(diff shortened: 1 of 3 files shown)


src/storage_manager/model/RemoteStorage.js
```diff
@@ -74,7 +74,7 @@ module.exports = require('backbone').Model.extend({
   },
 
   load(keys, clb) {
-    this.request(this.get('urlLoad'), {body: {keys}}, clb);
+    this.request(this.get('urlLoad'), {body: {keys}, method: 'get'}, clb);
```
santiph: Body on GET?? @.@

artf:
One of the ideas of `urlLoad` is too get only requested resources by using `keys` (eg. `["html", "css", ...]`). Unfortunately, the fetch specs don't allow an 'easy' way to add query parameters (https://github.com/github/fetch/issues/256) so this is why I left it as it is 😬
So before merging, I'd say we need to transform that array in a query string (eg. `keys[]=val&key[]=val2`)

santiph:
What about a comma separated list of keys? `/templates/42?keys=assets,css,styles,html,components`

artf:
What do you think about removing `keys` completely? If anybody need to fetch one resource or more all you have to do is to update `urlLoad` with your `keys` and logic.

I don't even see the point of adding `bodilessMethods` check, grapesjs's fetch polyfill was made to fit its own needs, if someone will use it for other scenarios it's not my responsability to check their requests, moreover you're applying the same logic twice (in `RemoteStorage.js` and `fetch.js`) :)

So, if you just remove the `keys`, from the `load` method, I'm gonna merge it

artf:
Thanks @santiph 👍

santiph:
Thanks to you, @artf  ;) 
Would you mind to publish a new version of grapesjs into npm?

artf:
published

PR #458, opened 2017-10-27
Add Arguments to Editor.trigger() function

santiph:
Arguments added to a triggered event are not being passed along to backbone's trigger function.


artf:
good catch thanks Santiago 👍

PR #465, opened 2017-10-30
Update index.html so that all media files have an absolute path

DeMongton:
When running the GrapesJS demo locally I noticed that some assets point to invalid paths. e.g. "/img/video2.webm" causing them not to load.

I also noticed that most media was getting loaded from the web, some of them directly from http://grapesjs.com

This pull request is to ensure that those assets that originally had an invalid path, will now be able to load from http://grapesjs.com

### DIFF
--- a/index.html
+++ b/index.html
@@ -89,7 +89,7 @@
       <section class="am-sect">
         <div class="container-width">
           <div class="am-container">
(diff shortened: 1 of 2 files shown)


artf:
great, thanks Pascal

PR #474, opened 2017-11-01
Add getConfig() function for SelectorManager

adamlc:
SelectorManager is missing a getConfig() function. This adds it in. Only really need it for translation purposes :)


artf:
Great, thanks Adam

PR #489, opened 2017-11-07
Fixing background color for gjs-input-unit

leantorres73:
Same issue as comboboxes background color for firefox.(https://github.com/artf/grapesjs/issues/456).

After the MR, the selects with class "gjs-input-unit" are still having this issue.

captura de pantalla 2017-11-07 a la s 16 18 46

Just a quick fix and include the fields option inside the scss


artf:
thanks @leantorres73

PR #496, opened 2017-11-08
add editable attribute to image component

ryandeba:
do not run the open-assets command when editable==false

### DIFF
--- a/src/dom_components/model/ComponentImage.js
+++ b/src/dom_components/model/ComponentImage.js
@@ -8,6 +8,7 @@ module.exports = Component.extend({
     src: '',
     void: 1,
     droppable: 0,
+    editable: 0,
     highlightable: 0,
     resizable: 1,
     traits: ['alt']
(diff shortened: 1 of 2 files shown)


src/dom_components/model/ComponentImage.js
```diff
@@ -8,6 +8,7 @@ module.exports = Component.extend({
     src: '',
     void: 1,
     droppable: 0,
+    editable: 1,
```
ryandeba: I used 1 here for consistency with the other properties. Just curious, but is there a reason that these are numeric instead of boolean? I know that 0 evaluates to false and 1 evaluates to true, so functionally it's fine...is it just personal preference?

artf: yeah, it's totally personal, using 1 character instead of 4/5 for the same functionality makes me feel better 😄

artf:
thanks @ryandeba

PR #506, opened 2017-11-11
split webpack.config.js into dev, prod and common versions

### LINKED ISSUE #433: Unminified grapjes.js file?
I downloaded the git repo but was only able to find grapes.min.js. I really want to dig into the code some more and get a better understanding of how all the files are related to each other but I'm not seeing the unminified version anywhere. Could someone point me to where it is in the repo or if not, could the unminified version be made available somewhere?

Yaevh:
I've split `webpack.config.js` into `dev`, `prod` and `common` files to get rid of that ugly `if(env != 'DEV')`. This also produces and unminified version of grapes.js on build and in dev server, which resolves #433.


artf:
Thanks for your contribution Tomasz. First of all, seems like you're pushing from a not updated branch (your webpack still has a jquery), always `git pull` before a PR.
I'd say that, **personally**, I agree about `if(env != 'DEV')`, but also find much more ugliest creating **3** different files for such a tiny configuration.
I close this for now, but would really like to see a solution for #433 without overwhelming stuff :)

PR #515, opened 2017-11-13
babel-polyfill added and promise-polyfill removed

ateshuseyin:
In this PR, I added babel polyfill into project. And removed promise-polyfill package from the project. Because babel-polyfill already includes promise polyfill inside. I think these changes resolves issue #214 .


artf:
Thanks @ateshuseyin

PR #517, opened 2017-11-13
surge config 

ateshuseyin:
surge is a static web publishing service for developer and most of feature is free. I mainly use surge to add live link into PRs as done before in #515. 

In this PR I added a small config file which prevents deployement of unnecessery files.

### DIFF
--- a/.surgeignore
+++ b/.surgeignore
@@ -1,0 +1,4 @@
+!dist
+node_modules
+src
+test


artf:
Thanks @ateshuseyin for the PR but I'd like to stop keep adding configuration files from dev's personal preferences, otherwise this repo will become a bucket of config files, this also confuses a lot of people (I'm one of them). If anyone wants to use it, I'd rather accept to add it to the `.gitignore`, but grapesjs don't need it

PR #533, opened 2017-11-16
Guard against the style object being a string

### LINKED ISSUE #532: Error when trying to hide some elements from the Layers Manager
Steps:
1. Go to http://grapesjs.com/demo.html
2. Open Layer Manager.
3. Click on the "Build your templates without coding" text element.
4. Click on the "eye" button in the Layer Manager to hide the element.

Result: 
The following exception is thrown: 
Uncaught TypeError: Cannot create property 'display' on string ''
and the element is not hidden.

I believe I found the cause and will create a pull request with a fix.

tomichal:
Fixes #532

### DIFF
--- a/src/navigator/view/ItemView.js
+++ b/src/navigator/view/ItemView.js
@@ -193,6 +193,8 @@ module.exports = Backbone.View.extend({
 
     var cCss = _.clone(this.model.get('style')),
     hClass = this.pfx + 'hide';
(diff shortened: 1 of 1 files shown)


artf:
@tomichal thanks for the PR, but already fixed here https://github.com/artf/grapesjs/releases/tag/v0.12.37

Issue #162, opened 2017-07-13
Position of dragged block


adrian1207:
Hi, I have a second problem, with start position of dragged block when there is some content above GrapesJS editor.

Is it a problem, becouse grapes needs some absolute div for container? Or some problem with scorlled site, when Grapes is below content?

Here is screen

![grapesdrag](https://user-images.githubusercontent.com/9296448/28159938-ecfe45ce-67be-11e7-8bf4-8701634b6aec.jpg)

PR #534, opened 2017-11-16
Clear cached scroll offset on window scroll

### LINKED ISSUE #495: Highlighter and toolbar in the wrong place
Hi,

From time to time, on my canvas the highlighter and toolbar appear in the wrong place. Is this a known bug? Do I have to manually refresh the canvas?

It seems that the `top` css property on `gjs-highlighter` and `gjs-badge` aren't being calculated properly, they still line up horizontally just not vertically. Any ideas?

I've attached screenshots below. 

Thanks,
T

screen shot 2017-11-08 at 16 44 30
screen shot 2017-11-08 at 16 44 51
screen shot 2017-11-08 at 16 45 05

tomichal:
Fixes #495 (a bug where the toolbar, highlighter, etc elements’ position 
would be off in editors positioned inside a scrollable window).

### DIFF
--- a/src/canvas/view/CanvasView.js
+++ b/src/canvas/view/CanvasView.js
@@ -5,6 +5,7 @@ module.exports = Backbone.View.extend({
 
   initialize(o) {
     _.bindAll(this, 'renderBody', 'onFrameScroll', 'clearOff');
+    window.onscroll = this.clearOff
     this.config = o.config || {};
     this.em = this.config.em || {};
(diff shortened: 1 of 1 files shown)


artf:
Great, seems to work perfectly @tomichal  :1st_place_medal: 

Would you mind to check also this https://github.com/artf/grapesjs/issues/162 😬  which is also affected by a scrolled page?

tomichal:
Just created a pull request for a fix for the #162 @artf :)

PR #539, opened 2017-11-17
Fix block component drag position issue

### LINKED ISSUE #162: Position of dragged block

tomichal:
Fixes #162


artf:
Thanks @tomichal, really appreciate your help 👍 
Anyway, I see that it fixes the starting position but still persists when you drag it over the canvas, isn't the same for you?!?

tomichal:
Ah, sorry @artf - I didn't account for a scroll within the editor canvas (I must've skipped a check for when the editor canvas was scrolled).

artf:
Thanks for the update @tomichal but I still see this (I use `margin-top` on `#gjs` container)
scroll

Don't know if it the same for you, can you confirm?

src/utils/Sorter.js
```diff
@@ -154,18 +153,26 @@ module.exports = Backbone.View.extend({
...
-      addTop = frameRect.top || 0;
-      addLeft = frameRect.left || 0;
-    }
-    var hStyle = this.dragHelper.style;
-    hStyle.left = (e.pageX - win.pageXOffset + addLeft) + 'px';
-    hStyle.top = (e.pageY - win.pageYOffset + addTop) + 'px';
+      // TODO: Cache ?!?
+      var frameRect = frame.getBoundingClientRect();
+      var ownerDocBodyRect = frame.ownerDocument.body.getBoundingClientRect();
+      addTop = frameRect.top - ownerDocBodyRect.top - doc.body.scrollTop;
```
artf: The `doc` in this case is the one from the canvas and its scroll is irrelevant for us, we need the main document's scroll. 
By the way, be aware, if you use `document.body.scrollTop` it won't gonna work (will return always `0`), use `document.documentElement.scrollTop` instead

tomichal: Hey @artf I'll check those issues... but I could swear the doc.body.scrollTop worked in that case, hmm...

tomichal: Here are my observations:
1. Re `document.body.scrollTop` - it worked for me (in Chrome) - it returns a scroll offset of the canvas inside the editor's iframe and I wanted that to offset the Y position returned from the `e.pageY` - since the `e` is an event coming from the iframe, the `pageY` is a value from the top of the iframe, but that will be too large if the iframe is scrolled, hence the use of `doc.body.scrollTop` to offset that? And `doc.documentElement.scrollTop` was always returning 0 for me, but `doc.body.scrollTop` was showing the correct scroll offset.
2. Re the bug shown in the screencast: Yup...  I noticed that in situations when the first element of the body has a margin set on it. If you place an element with a height set on it, rather than a margin, the positions will appear correctly (even if then you also have a top margin set on the `#gjs` container).

artf: > 1. Re document.body.scrollTop - it worked for me (in Chrome) - it returns a scroll offset of the canvas inside the editor's iframe...

Did you try then to drag inside a scrolled canvas, because I'm facing issues when taking into account canvas's offset. For me, it seems to work with something like this

```js
var posY = e.pageY;
...
if (frame) {
  posY = e.clientY;
  var frameRect = frame.getBoundingClientRect();
  addTop = frameRect.top + document.documentElement.scrollTop;
  ...
}
...
dragHelperStyle.top = (posY + addTop) + 'px';
```

tomichal: Yes - I dragged inside a scrolled canvas, as well as the scrolled container. Odd... well if your code works, then just roll with it?
Just to confirm, although no reason for the git diff to be wrong, but this is the full code of that function on my end:

    moveDragHelper(e) {
      var doc = e.target.ownerDocument;
  
      if(!this.dragHelper || !doc) {
        return;
      }
  
      var addTop = 0;
      var addLeft = 0;
      var window = doc.defaultView || doc.parentWindow;
      var frame = window.frameElement;
      var dragHelperStyle = this.dragHelper.style;
  
      // If frame is present that means mouse has moved over the editor's canvas,
      // which is rendered inside the iframe and the mouse move event comes from
      // the iframe, not the parent window. Mouse position relative to the frame's
      // parent window needs to account for the frame's position relative to the
      // parent window.
      if(frame) {
        // TODO: Cache ?!?
        var frameRect = frame.getBoundingClientRect();
        var ownerDocBodyRect = frame.ownerDocument.body.getBoundingClientRect();
        addTop = frameRect.top - ownerDocBodyRect.top - doc.body.scrollTop;
        addLeft = frameRect.left - ownerDocBodyRect.left - doc.body.scrollLeft;
      }
      dragHelperStyle.top = (e.pageY + addTop) + 'px';
      dragHelperStyle.left = (e.pageX + addLeft) + 'px';
    },

tomichal: Oh, not sure if that would be a factor but I was testing it against the v0.12.37 tag rather than the dev branch...

tomichal: Ok, just pulled the latest dev branch and tests on there look fine for me (puzzled).

artf: Hi @tomichal I confirm that your code works correctly without margins at first element, but mine fix seems to work good in both cases. I merge your PR and then will apply my fix.
Thanks for your time

tomichal: Sounds good :) ... and btw, thank you for creating this editor - I think it’s an awesome tool.

PR #547, opened 2017-11-20
updated package.json and webpack.config.js to enable unminified build

### LINKED ISSUE #433: Unminified grapjes.js file?

Yaevh:
Another take on #506 - this one uses webpack's own `--env` instead of `cross-env WEBPACK_ENV` to produce both minified and unminified output file (or unminified only, if using `npm run build-dev`). Resolves #433.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,11 +74,12 @@
   },
   "scripts": {
     "lint": "eslint src",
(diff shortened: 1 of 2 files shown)


artf:
Thanks mate 👍

PR #550, opened 2017-11-20
Include value-less attributes in component.toHTML()

ryandeba:
This is to fix an issue where boolean attributes (http://w3c.github.io/html/infrastructure.html#sec-boolean-attributes) are being excluded in the HTML returned from editor.getHtml(). For example, starting with code that looks like `<input type="checkbox" checked>`, I would expect the the "checked" attribute to be present when calling editor.getHtml(), but it currently is not. I believe this behavior should apply to custom attributes as well (data-*). 

According to the spec (linked above) it is valid to output boolean attributes with an empty string value, so I think these changes will work for standard boolean attributes (`checked`, `selected`, `disabled`) as well as custom attributes that may or may not have a value (`data-boolean` / `data-test=""`).

### DIFF
--- a/src/dom_components/model/Component.js
+++ b/src/dom_components/model/Component.js
@@ -554,7 +554,7 @@ module.exports = Backbone.Model.extend(Styleable).extend({
     for (let attr in attributes) {
       const value = attributes[attr];
 
-      if (!isUndefined(value) && value !== '') {
+      if (!isUndefined(value)) {
           attrs.push(`${attr}="${value}"`);
       }
     }
--- a/test/specs/dom_components/model/Component.js
+++ b/test/specs/dom_components/model/Component.js
@@ -82,6 +82,16 @@ module.exports = {
           expect(obj.toHTML()).toEqual('<article data-test1="value1" data-test2="value2"></article>');
         });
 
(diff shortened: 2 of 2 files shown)


artf:
`value !== ''` was here to keep html clean in case one removes a value of the attribute, but not having boolean attributes it's actually a big issue, so for now I merge it, thanks 👍

ryandeba:
That's a good point about removing the value from an attribute... not sure what the right solution for that is other than just removing the attribute rather than clearing it's value.

I was trying to think of a way to distinguish between boolean attributes vs. attributes that have an empty string value, but couldn't think of a good solution. There's not really a good way to programmatically add a boolean attribute right now either, other than setting it's value to an empty string. Maybe boolean attributes should be stored separately from non-boolean attributes? I don't know...feel free to share any thoughts you have on how to improve this and I can take another pass at it if you want.

Anyways, thanks for the merge!

artf:
> There's not really a good way to programmatically add a boolean attribute right now either, other than setting it's value to an empty string.

Yeah, it's exactly what I thought and probably the best solution would be to provide a way to remove the attribute, but currently, I don't see it as something strictly indispensable

PR #560, opened 2017-11-22
fix bug with deleting from asset manager

ryandeba:


src/asset_manager/view/AssetImageView.js
```diff
@@ -79,7 +79,7 @@ module.exports = require('./AssetView').extend({
    * @private
    * */
   onRemove(e) {
-    e.stopPropagation();
+    e.stopImmediatePropagation();
```
ryandeba: I thought that e.stopPropagation() would work, but found this after a quick google search: https://stackoverflow.com/questions/5457915/how-can-i-stop-event-propagation-with-backbone-js

src/asset_manager/view/AssetImageView.js
```diff
@@ -1,9 +1,9 @@
 module.exports = require('./AssetView').extend({
 
   events: {
+    'click [data-toggle=asset-remove]': 'onRemove',
```
ryandeba: I couldn't find it documented anywhere, but these events seem to fire in the order they are registered. The onRemove function needs to run before onClick so that the event propagation can be stopped.

ryandeba:
This is the fix https://github.com/artf/grapesjs/issues/549

src/asset_manager/view/AssetImageView.js
artf: Great, this is actually working 👍

artf:
hey Ryan ping me on discord chat pls

PR #561, opened 2017-11-22
add hoverable property to component objects

ryandeba:


ryandeba:
This is for the hoverable property discussed here: https://github.com/artf/grapesjs/issues/554

PR #570, opened 2017-11-24
render text view after disabling editing

ryandeba:
this is to fix a bug where text content can appear to be duplicated after adding text content to a component that originally had no content

https://github.com/artf/grapesjs/issues/569

### DIFF
--- a/src/dom_components/view/ComponentTextView.js
+++ b/src/dom_components/view/ComponentTextView.js
@@ -82,7 +82,7 @@ module.exports = ComponentView.extend({
         }
 
         // Avoid re-render on reset with silent option
-        model.set('content', '');
+        model.set('content', '').trigger('change:content', model);
         comps.add(content);
         comps.each(model => clean(model));
         comps.trigger('resetNavigator');
@@ -91,7 +91,6 @@ module.exports = ComponentView.extend({
 
     this.rteEnabled = 0;
     this.toggleEvents();
-    this.render();
   },
 
   /**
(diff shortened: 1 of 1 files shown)


src/dom_components/view/ComponentTextView.js
```diff
@@ -91,6 +91,7 @@ module.exports = ComponentView.extend({
 
     this.rteEnabled = 0;
     this.toggleEvents();
+    this.render();
```
ryandeba: I saw your comment above about "double content" (lines 67-68)...seems like it's related, so maybe there's a place to fix this?

artf: Unfortunately, this breaks stuff, try for example to remove a word and then execute undo and redo, this will duplicate the content. The issue is with `model.set('content', '')` (line 85), it doesn't trigger the change as the `content` is already empty so it doesn't reset its content. 
For now the only fix I see it's something like this `model.set('content', Math.random()).set('content', '')`

src/dom_components/view/ComponentTextView.js
```diff
@@ -82,7 +82,7 @@ module.exports = ComponentView.extend({
         }
 
         // Avoid re-render on reset with silent option
-        model.set('content', '');
+        model.set('content', '').trigger('change:content', model);
```
ryandeba: @artf What do you think about explicitly triggering the `change:content` event instead of setting it to a random value? I think this a bit more clear what's going on, but I'd be happy to change it like your suggestion if you prefer.

artf: Seems definitely less hacky 😬

artf:
Thanks Ryan 💜

Issue #583, opened 2017-11-28
Event stop:preview doesn't fire 


gabrigcl:
Hi. I need the firing of this event in this situation: I have some elements out of the canvas that need to be hide at preview mode and shown when I exit preview mode. However, the event stop:preview doesn't fire when I exit the preview and turn back to canvas editing. I've made a solution myself that works well, but I don't know if this is an appropriate solution. That's why I'm sending a screenshot of my solution from my private git instead of sending a pull request, and I hope you could implement this on grapes if it is a good solution. Thanks.
![captura de tela 2017-11-28 as 16 27 05](https://user-images.githubusercontent.com/7770605/33340655-76e63d7e-d45b-11e7-86ea-c0685d6aeba2.png)

theanhmx:
@artf sorry for this stupid question, 
but when will the fix available in the npm package?
or from which branch i can manually pull the code with this fix and build?

PR #592, opened 2017-11-29
Problem event stop:preview not firing solved

gabrigcl:
The related event (stop:preview) from src/commands/view/Preview.js were not firing (issue #583). I've found this error in this situation:

I have some elements out of the canvas that need to be hide at preview mode and shown when I exit preview mode. However, the event stop:preview doesn't fire when I exit the preview and turn back to canvas editing. I've made a solution myself that works well, look the commit. Here a screenshot: 
captura de tela 2017-11-28 as 16 27 05


artf:
Great 👍  Thank you @gabrigcl

PR #636, opened 2017-12-13
Bug in Modal while trying to use it with the video component solved

gabrigcl:
Hi. Recently working with my project with grapesjs I've found this bug. Here is the situation: I wish to use the Modal window provided by grapes to load and configure the traits of the Video component. My goal is, as soon as I drop the video component on the canvas, the modal window must appear automatically, with the video traits for setting its properties. To achieve this, I had to set "activeOnRender: 1" in the video block I've created, so that the video plugin I've created too, listen to the event "active" when I drop the component on the canvas.

However, I just set "activeOnRender: 1" in the video block and the modal window  appear as soon as I drop the video component on the canvas, and, for my surprise, the title and content of Image Component modal (assets manager) is loaded on the modal window (and I didn't even listened to the event "active" in the my video plugin). And no matter if I listen to the event "active" in my video plugin, the problem persists, the Image modal keeps showing for the video.

I discovered that this is happening because of the heritage mechanism used for the native video component, that inherits from the Image component. It's just a javascript context problem. You can see my solution on my commit. Now I'm able to continue the development of my plugin, with the blank modal available for me as I drop the video component, and I believe my solution didn't break any functionality of the framework.


artf:
Seems legit, thanks @gabrigcl

PR #652, opened 2017-12-15
Add colorPicker config options

ryandeba:
https://github.com/artf/grapesjs/issues/616

### DIFF
--- a/src/domain_abstract/ui/Input.js
+++ b/src/domain_abstract/ui/Input.js
@@ -25,7 +25,7 @@ module.exports = Backbone.View.extend({
     const ppfx = opts.ppfx || '';
     this.opts = opts;
     this.ppfx = ppfx;
-    this.target = opts.target || {};
+    this.em = opts.target || {};
     this.listenTo(this.model, 'change:value', this.handleModelChange);
   },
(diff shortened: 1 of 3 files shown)


ryandeba:
Let me know if you want me to rename more `target` variables to `em`...not sure how far up the chain you were wanting to take that.

Prior to making my changes I was getting some errors when trying to add a "color" trait to a component type. I don't have any uses for trait colors at the moment, but I thought I would mention it in case you were unaware (but maybe I'm just doing something really wrong?). Here's my code and the error that I was seeing, let me know if you want a live example:
```
var textType = editor.DomComponents.getType("text");
textType.model.prototype.defaults.traits.push({ // add a "color" trait for testing purposes
	type: "color",
	label: "color",
	name: "color"
});
```
Interacting with the trait color picker resulted in this error: `Uncaught TypeError: model.setValueFromInput is not a function`

artf:
Good job, thanks Ryan 👍 
I checked the error and it's actually true, InputColor expects the model to be a `StyleManager/Property` which is obviously not the same for traits 😬  have to refactor that part, thanks for the clue

PR #666, opened 2017-12-21
skip non-selectable parents in select-parent command

ryandeba:
https://github.com/artf/grapesjs/issues/660


artf:
Great, thanks Ryan  🎖 👍

PR #671, opened 2017-12-22
When removing all classes from a component the model was not reset

mingodad:
This fix the problem mentioned in https://github.com/artf/grapesjs/issues/661


artf:
Thanks for the PR Domingo but as already mentioned [here](https://github.com/artf/grapesjs/issues/661#issuecomment-353446941) the real issue was different and already fixed. The patch will be available in the next release

PR #676, opened 2017-12-23
Escape html

mingodad:
When exporting/importing html we need to properly escape/unescape then.

Example:
 <p>A &lt;div&gt; tag can be used for ...</p> ... <input type="text" placeholder="class=&quote;cl&quote;">


artf:
Hi @mingodad, unfortunately, I can't merge this as it will break all other imports/exports with current text elements (which should not be escaped). I think we just have to create a new Component for `<pre>` tags which escapes its contents.

ps. I'd prefer to use `escape` from `underscore`
```js
import { escape } from `underscore`;

const escaped = escape('string');
```

PR #685, opened 2017-12-26
Remove duplicated dblclick event listener

ryandeba:
Hi @artf,

This is just a small change to reduce some duplicated code. The `'dblclick': 'enableEditing'` event is inherited from ComponentTextView, so it seems unnecessary to repeat that code on the ComponentLinkView object. I was experimenting with overriding the dblclick behavior of the ComponentTextView and realized that my changes were not getting applied to links.


artf:
Thanks Ryan 👍

Issue #668, opened 2017-12-22
When element not available to select, suggest hover and click event will select the available parent.


tommywulove:
Hi @artf, can i have a enhancement for select command, when the cursor click or hover the element which is not selectable, then the selection jump to available element. Since the element not available to select, and nothing can do in the element, so i suggest hover and select jump to available parent.

tommywulove:
Hi @artf ,
I am trying to help for this enhancement. But i am not sure am i correct, please help to verify.

PR #691, opened 2017-12-28
#668 enhancement

tommywulove:


src/commands/view/SelectComponent.js
```diff
@@ -100,7 +100,21 @@ module.exports = {
       this.onFrameScroll(e);
       this.updateAttached();
     }
+    var model = $(trg).data('model');
+    if (typeof model != 'undefined') {
 
+      if (!model.get("selectable")) {
+        var comp = model && model.parent();
```
artf: In this case, I think it makes more sense checking against `hoverable` property

src/commands/view/SelectComponent.js
```diff
@@ -100,7 +100,21 @@ module.exports = {
       this.onFrameScroll(e);
       this.updateAttached();
     }
+    var model = $(trg).data('model');
+    if (typeof model != 'undefined') {
```
artf: Just use `if (model) {`

src/commands/view/SelectComponent.js
tommywulove: updated.

src/commands/view/SelectComponent.js
tommywulove: updated.

PR #702, opened 2018-01-03
add "typeof" to fix bug with variable inspection

ryandeba:
This needs to be `model != undefined` or `typeof model != 'undefined'`, as the `model` variable should never be the string "undefined". The code as currently written throws an error when model is undefined.


artf:
👍

Issue #679, opened 2017-12-23
Categorize Traits similar to Blocks


z1lk:
I'm building the Bootstrap plugin and have a growing list of utility settings for basically every component. Add component-specific and ARIA settings on top of that, and the Settings panel becomes pretty cramped. Can we have the ability to categorize settings into "collapses" like with Blocks?

I'd be willing to contribute a PR if we can do this by simply copying the Blocks categories code.

artf:
Thanks @z1lk for the suggestion, it definitely makes sense but I'm not sure about copying stuff, I'd like to avoid repeating code, probably it would make sense create an abstract view which does the same thing and which we can use in other points

artf:
Closing this in favor of https://github.com/GrapesJS/grapesjs/issues/1806

PR #728, opened 2018-01-08
Trait categories

z1lk:
Implements changed requested in #679. Took me some time after the holidays to get to this. I also have only an elementary understanding of Backbone so please instruct me on good convention/patterns if you'd like! This was a learning experience for me, so also don't hesitate to tell me if something looks utterly wrong.

Changes:
- moved existing Category model/view to `domain_abstract`. Is this a folder for general abstractions?
- created abstract Categorizable model, mostly taken out of the Block model, which creates that Category during initialization. I'm not sure if this is necessary since Category creation also happens in the view.
- created abstract CategorizableView, mostly taken out of the BlocksView, which handles rendering of the cats/no-cats containers. This extends DomainViews because TraitsView was extending that. Notice the new DomainViews method `getItemView`, which selects either the itemView or one of the itemsView. This method is called in the new CategorizableView's `add` method. Also notice the `categorizableType` property being passed down into this view. This is because "block" was hardcoded into the container class names.

There's a file `index-with-categorized-traits` that you can copy to `_index.html`, which has blocks and traits setup for testing. Let me know if this is the proper direction, and I'll make changes if needed, clean the code up, and look through the tests.


artf:
hey thanks @z1lk but before the review, can you update tests first?

PR #747, opened 2018-01-11
Categories order

JuanTincho:
- Add a new `order` style ( if exists) in the CategoryView component.
- Use flexbox style to class `.gjs-block-categories`.


artf:
Great 👍 , thanks Juan

PR #749, opened 2018-01-11
set the initial trait color input value after the InputColor is created

ryandeba:


src/trait_manager/view/TraitColorView.js
```diff
@@ -10,16 +10,17 @@ module.exports = TraitView.extend({
...
         contClass: this.ppfx + 'field-color',
         model: this.model,
         ppfx: this.ppfx
       });
-      this.input = inputNumber.render();
+      this.input = inputColor.render();
       this.$input = this.input.colorEl;
       value = value || '';
       this.model.set('value', value).trigger('change:value');
+      this.input.setValue(value);
```
ryandeba: Hey @artf,

Small followup from my last pull request. I just noticed that the input color value is always initially "undefined" after the InputColor is created. Setting the value here into the input here seems to resolve it.

artf: Good catch 👍

PR #752, opened 2018-01-12
fix sunset at the left edge of the window

MuTaToR08:
If the element is small and is close to the left edge of the screen, the toolbar goes over the edge and is not visible.


artf:
Thanks @MuTaToR08 👍

PR #797, opened 2018-01-25
Feature/updated npm

Piterden:


artf:
Hi Denis, can you leave the webpack file as it is?

Piterden:
Ya, sure

Piterden:
Like this?

artf:
Thanks Denis, but one more thing. Why do we need [`babel-eslint`](https://github.com/babel/babel-eslint#why-use-babel-eslint)?
When you push a PR you should avoid the `build` command otherwise we get useless conflicts with `dist` files

artf:
Update deps with the latest version

PR #798, opened 2018-01-25
Feature/codestyle es6

Piterden:


artf:
@Piterden 
please void changing the styleguide of the project especially without discussing such a thing
https://github.com/artf/grapesjs/blob/dev/CONTRIBUTING.md#styleguide

Issue #796, opened 2018-01-25
Can't drag component into canvas after styled element is cloned


kickbk:
Go to any of the demo, select a text box, give it a background color. Clone it. Then try to drag in another text component and it doesn't let you.

artf:
@kickbk I can't reproduce it, but I think I'm doing something wrong, can you provide a screencast/gif of the behavior? (or anyone who faces the same issue)
Thanks

kickbk:
![2018-01-25 09 44 28](https://user-images.githubusercontent.com/31323376/35403505-c4a25314-01b4-11e8-939c-12ab86614813.gif)

artf:
Thanks @kickbk now I get it. It was already fixed here #779 and will be available in the next release

PR #799, opened 2018-01-25
Fix for issue #796

cjpollard:
Proposed fix for https://github.com/artf/grapesjs/issues/796 .  I've bound drag events to the picker element rather than the document, and unbound from the same.


artf:
Thanks @cjpollard BTW it was already solved https://github.com/artf/grapesjs/issues/796#issuecomment-360610181

PR #817, opened 2018-01-29
Modifications in '.cm-s-hopscotch span.cm-error'

alimajed93:
Hi @artf ,

I noticed a bug, which is when the user types in the modal "<<" (2 consecutive "<"), or when the user wants to close an opened html tag (`</div>`), the background and the font go both red, which disables the user from seeing what is being typed.

I noticed that the property `color: #989498` for `.cm-s-hopscotch span.cm-error` is being overridden. So I made the following changes int the file /dist/css/grapes.min.css:

1. I moved `.cm-s-hopscotch span.cm-error{}` forward so they are not overridden anymore

2. I changes the color from #989498 to while (#ffffff) for better readability

The following 3 images show the initial state, the fix with dark grey color and the final result.
screen shot 2018-01-29 at 4 38 13 pm
screen shot 2018-01-29 at 4 37 32 pm
screen shot 2018-01-29 at 4 40 05 pm

Thank you!


artf:
Good point @alimajed93 but can you update here the file https://github.com/artf/grapesjs/blob/dev/src/styles/scss/main.scss#L629
Files inside `dist`(distribution) are just generated on `npm run build` command and shouldn't be touched

alimajed93:
Okay @artf .
Modifications done.

artf:
Thanks @alimajed93 👍

PR #819, opened 2018-01-29
Added resize event to the toggleSelectComponent function.

JuanTincho:
Acording to issue: https://github.com/artf/grapesjs/issues/810 and https://github.com/artf/grapesjs/issues/699


artf:
Thanks Juan 👍

PR #870, opened 2018-02-14
Hide very large labels on blocks

JuanTincho:
* Added css to hide very large labels. 
* Added title to label class, thus it shows the text on a tooltip.

Before:

image

Now:

label tooltip


artf:
Cool idea. Thanks Juan

Issue #914, opened 2018-02-28
[BUG]: Grapes cannot be reinitiated with new HTML


nojacko:
Calling `init()` on Grapes after changing HTML fails. This kinda situation happens on one page apps.

I think a `destory()` function to gracefully reset Grapes is needed.

1. https://jsfiddle.net/wdtrg1je/ (very bare bones simulation of a one page app)
1. Click "Add Grapes" (loads Grapes with a some HTML)
1. Click the "Link", change it's href to `http://yahoo.com`
1. Click "Remove Grapes". This grabs the HTML from Grapes before destorying the grapes element.
1. Click "Add Grapes" again. This starts Grapes again with the new HTML which differs from the original HTML.
1. Error in console.

![screen shot 2018-02-28 at 14 54 20](https://user-images.githubusercontent.com/1667141/36794589-aba3667e-1c98-11e8-881a-8a66136f4fa2.png)

If "Add Grapes" and "Remove Grapes" are clicked without any changes to HTML, it's fine.

artf:
As already mentioned [here](https://github.com/artf/grapesjs/issues/849#issuecomment-364279629), currently, I don't see any value in starting such a feature (creating a good destroy method) but would really happy to see some PR :)

nojacko:
Looks like I'm not the only one who wants this 😉 . Single page apps are where this has value (the issue I raise but also freeing up some memory).

I'll look to raise a PR when I got something.

PR #920, opened 2018-03-02
Fixed TypeError when canvas has been removed.

nojacko:
For issue #914. 

This occurs when something in grapes is selected, grapes' canvas removed and grapes re-init'd.

Fixed just checks the element exists first and returns null if it doesn't.


artf:
Thanks James 👍

PR #928, opened 2018-03-06
prevent grapes.js from being injected twice during development

tommedema:
Very minor fix on development where previously the javascript bundle was injected twice:

screen shot 2018-03-06 at 09 44 54

After this PR this is fixed:

screen shot 2018-03-06 at 09 45 18

Does not affect tests.


artf:
Thanks Tom

PR #929, opened 2018-03-07
show message in trait manager panel when there is no comp selected

ryandeba:
This is for https://github.com/artf/grapesjs/issues/869


artf:
Thanks Ryan 👍 👍 👍

PR #934, opened 2018-03-08
Escape attributes in Component.toHTML

ryandeba:
This is for https://github.com/artf/grapesjs/issues/912

Rather than manually escaping/encoding certain characters in the attributes, I thought it would be better to let the browser do that for me by returning the `outerHTML` of a populated element.

### DIFF
--- a/src/dom_components/model/Component.js
+++ b/src/dom_components/model/Component.js
@@ -669,20 +669,33 @@ const Component = Backbone.Model.extend(Styleable).extend(
       const sTag = model.get('void');
       const attributes = this.getAttrToHTML();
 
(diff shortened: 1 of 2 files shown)


artf:
Thanks for the PR mate, but unfortunately I have to refuse this one. Adding such a change not only will introduce an impact in performances but we'll probably get a lot of inconsistencies (each browser might have its own rules for managing elements and "cut" some expected custom component). For this issue, I'd simply and trivially escape double quotes

ryandeba:
No worries @artf! I'll plan to take another pass at this by just escaping the double quotes later today or tomorrow.

PR #971, opened 2018-03-21
Remove unnecessary duplicated property in block_manager config.js

chapterjason:
Hey,
as you can see the configuration already has the property `appendTo` so I removed the duplicated property.


artf:
Thanks Jason

PR #988, opened 2018-03-29
Added style containers to fix styles/media-query order

vrudikov:
@artf This is the fix for style container we discussed in [896](https://github.com/artf/grapesjs/issues/896). Waiting for your comments


src/css_composer/view/CssRulesView.js
```diff
@@ -61,19 +62,41 @@ module.exports = require('backbone').View.extend({
...
+        let $stylesContainer = this.$el.find(`#${styleBlockId}`);
+        $stylesContainer.append(rendered);
+      }
     }
 
     return rendered;
   },
 
+  getMediaWidth(mediaText) {
+    return mediaText && mediaText.replace('(max-width: ', '').replace(')', '');
```
artf: `(max-width: ` it's not always the same, it depends on [`mediaCondition ` option](https://github.com/artf/grapesjs/blob/dev/src/editor/config/config.js#L91)

src/css_composer/view/CssRulesView.js
```diff
@@ -61,19 +62,41 @@ module.exports = require('backbone').View.extend({
...
+    // Create devices related DOM structure
+    const pfx = this.pfx;
+    this.em
+      .get('DeviceManager')
+      .getAll()
+      .forEach(function(model) {
+        const blockId = pfx + 'rules-' + model.get('widthMedia');
+        $(`<div id="${blockId}"></div>`).appendTo(frag);
+      });
+
```
artf: It will work only when devices are already sorted by `widthMedia` which is not always true. Might be a case to use `sortBy` method:
```js
.getAll()
.sortBy(device => ...)
.forEach
```

src/css_composer/view/CssRulesView.js
vrudikov: Can I somehow simplify it? Do I have an option to take the current media condition parameter from the model directly - without any replacements?

artf: You should be able to get it directly from the config `this.em.getConfig(). mediaCondition`

vrudikov:
@artf Ready for review #2

vrudikov:
@artf Can you help me with the last failing test, plz. So the background... we have 2 tests and they are very similar: `Init editor from element` and `Adds a new storage and fetch correctly data from it`. Both of them share similar parts of code:
```
...
fixture.innerHTML = documentEl;
...
config.fromElement = 1;
...
```

And we have code in `CssRulesView.js.addToCollection` method where i build `styleBlockId`:
```
const styleBlockId = `#${this.pfx}rules-${mediaWidth}`;
```

When `Init editor from element` runs `this.pfx` is equal to `gjs-css-` and that is right behavior. 
But when `Adds a new storage and fetch correctly data from it` runs it seems that `CssRulesView` instance is loosing it's parent content and `this.pfx` is equal to just `css-`

That's why we get `Cannot read property 'appendChild' of null` for the line
```
fragment.querySelector(styleBlockId).appendChild(rendered);
```

artf:
I think the reason is that test asks for `storables` in EditorModel (due to `autoload`) which is passed by reference to its `defaults`. So a quick fix would be to add `this.set('storables', []);` [here](https://github.com/artf/grapesjs/blob/dev/src/editor/model/Editor.js#L54)

src/css_composer/view/CssRulesView.js
```diff
@@ -61,19 +62,50 @@ module.exports = require('backbone').View.extend({
...
+      }
     }
 
     return rendered;
   },
 
+  getMediaWidth(mediaText) {
+    return (
+      mediaText &&
+      mediaText.replace(`(${this.config.mediaCondition}: `, '').replace(')', '')
```
artf: mmm there is no `mediaCondition` inside `this.config`, you should get it from `this.em.getConfig('mediaCondition')`

vrudikov:
@artf Done. Please review

artf:
Thank you very much Valentin for this PR ❤️

vrudikov:
Now we just need to wait for the community tests :)

artf:
I'm quite confident to release it in the next version :)

PR #1010, opened 2018-04-05
implemented removeEmptyTextNodes boolean, fixes #989

### LINKED ISSUE #989: [BUG] Consecutive spans render whitespace differently in grapes' editor than in ordinary window
I've spent the greater part of an afternoon trying to debug this. While I have reduced this to a small and easy to reproduce test case, I didn't find the root cause of this issue yet. Hopefully someone can help.

Note the following HTML template:

```
<!DOCTYPE html>
<html>
<head>

<style>
body {
  margin: 0;
}

.navbar-main .main-navigation__list {
  margin: 0;
  padding: 0; 
}

.navbar-main .main-navigation__list .main-navigation__list-item {
  display: inline-block;
}
</style>
</head>

<body>
  <nav class="navbar-main">
    <div>
      <ol class="main-navigation__list">
        <li class="main-navigation__list-item">
          <span>Our solution</span>
        </li>
        <li class="main-navigation__list-item">
          <span>About</span>
        </li>
      </ol>
    </div>
  </nav>
</body>
</html>
```

If you render this in a browser, it renders as:

screen shot 2018-03-29 at 1 36 34 pm

Notice the whitespace between "Our solution" and "About".

If you then use the exact HTML and style as Grapes's template, it will render as:

screen shot 2018-03-29 at 1 36 44 pm

Notice that the whitespace is now not visible. The expected behavior is for grapes to render the template in the same manner as a browser does without grapes.

JSFiddle to reproduce:

https://jsfiddle.net/szLp8h4n/89/

Change `autorender: 0` to `autorender: 1` to trigger the bug.

Could this be because Grapes has a manual HTML and CSS parser, meaning that the DOM or CSS rules are replicated invalidly in this particular case? Can you figure out where it goes wrong?

tommedema:
Defaults to true to prevent breaking changes

Fixes #989

@artf please review, and if OK, merge and publish to NPM. :)

Cheers

### DIFF
--- a/src/editor/config/config.js
+++ b/src/editor/config/config.js
@@ -62,6 +62,9 @@ module.exports = {
   // Ending tag for variable inside scripts in Components
   tagVarEnd: ' ]}',
 
+  // Remove empty text nodes when parsed, unless they contain a space
+  removeEmptyTextNodes: true,
+
   // Return JS of components inside HTML from 'editor.getHtml()'
   jsInHtml: true,
 
--- a/src/editor/index.js
+++ b/src/editor/index.js
@@ -80,6 +80,7 @@
  * @param {Object} [config.domComponents={}] Components configuration, see the relative documentation
  * @param {Object} [config.panels={}] Panels configuration, see the relative documentation
  * @param {Object} [config.showDevices=true] If true render a select of available devices inside style manager panel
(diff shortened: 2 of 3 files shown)


artf:
@tommedema the PR is good thanks but I had to revert it as you put it on `master` instead of `dev`. Can you resubmit it on `dev` please :)

tommedema:
@artf resubmitted, please have a look at #1023

Issue #985, opened 2018-03-28
Can GrapesJS allow uploading arbitrary file through the built in asset manager?


tomichal:
I looked at the asset manager code as well as the wiki and it seems like despite the docs claiming that the asset manager can handle uploading arbitrary files (i.e. by adding an arbitrary asset type), the actual asset manager code hardcodes file type `image/*` on the file input field of the file upload form https://github.com/artf/grapesjs/blob/d5f0ae94dfd801a8c13cc0f6ce33e209f98f1c65/src/asset_manager/view/FileUploader.js#L8

Is this a bug/oversight and indeed the asset manager code needs to be modified/hacked, or am I missing a way to configure the manager to accept the file types matching the default asset types (image only so far?) as well as any arbitrary asset types?

My use case is to hook up the link component with the asset manager such that a user can upload a document file (pdf, doc, xls, ...) and upon doing so the asset manager updates the `href` attribute of the targeted link component.

artf:
Good question, I think the issue is the [OpenAssets](https://github.com/artf/grapesjs/blob/dev/src/commands/view/OpenAssets.js#L14) command itself that should be renamed in `OpenImageAssets` (indeed it renders only images types). So, what you can try to do is to create another command (eg. a more generic one or just for documents) and see if the [customization](https://github.com/artf/grapesjs/wiki/Assets#customization) works out for you.
If you will see any point of improvement in the process I'll be glad to discuss it.

tomichal:
@artf I created the referenced pull request. I admit my motivation was my specific use case, but hopefully it was coded generically enough to be acceptable in the core editor codebase.

tomichal:
... oh and just to clarify - another chunk of code written alongside this pull request resides in a custom plugin of my project. It's a plugin that essentially does the following:
1. Customises the Link DOM **component** such that upon creating a link via the RTE, the asset manager opens up at which point an arbitrary asset can be uploaded and linked to the newly created link component.
2. Customises the Image **asset type** of the asset manager such that upon selection of an image in the asset manager, either the `src` or `href` attribute of the target component get updated.
3. Adds an Other **asset type** to the asset manager. The Other asset type is used alongside the Image asset type if the asset manager is used for the Link DOM **component**.

PR #1028, opened 2018-04-10
Refactor ‘open-assets’ editor command to permit arbitrary file types

tomichal:
The file types can be passed as a config option for an asset type in which case the types will be used to limit the accepted file type when uploading new files in the asset manager.
(See #985 for motivation for this change).


tomichal:
Sounds good @artf , I'll update the pull request with the change you mentioned.

artf:
> you don't feel like defining the type of accepted files should belong in the definition of the asset manager asset type?

Exactly, for instance, this will allow showing one kind of type but accepting a more generic one

tomichal:
Made changes per your suggestion @artf (sorry it took so long)

artf:
Thanks @tomichal 👍

PR #1036, opened 2018-04-12
Allow multiple at-rules, fixes #1032

### LINKED ISSUE #1032: [BUG] Multiple font-face at-rule CSS rules are not processed
Latest version of grapes  #cfde1c1

If you supply a single `@font-face`, this is parsed just fine:

JSFiddle example: https://jsfiddle.net/szLp8h4n/140/

Screenshot:

screen shot 2018-04-11 at 3 19 56 pm

However, if you supply a second `@font-face`, only the second one is parsed, and the first is lost:

JSFiddle example: https://jsfiddle.net/szLp8h4n/141/

Screenshot:

screen shot 2018-04-11 at 3 21 45 pm

I am trying to debug this but would appreciate your help. Could it be related to [line 122 in ParserCss.js](https://github.com/artf/grapesjs/blob/dev/src/parser/model/ParserCss.js#L122)?

tommedema:
The CSS parser is trying to be smart by not adding CSS rules twice if they share the same selector. However, this does not work for at-rules, because these can be defined multiple times with the same selector, yet have different impact (e.g. multiple `@font-face` declarations).

- Fixes #1032 
- Added 3 more test cases
- All tests pass

### DIFF
--- a/src/css_composer/index.js
+++ b/src/css_composer/index.js
@@ -182,8 +182,13 @@ module.exports = () => {
       var w = width || '';
       var opt = { ...opts };
       var rule = this.get(selectors, s, w, opt);
-      if (rule) return rule;
-      else {
+
+      // do not create rules that were found before
+      // unless this is an at-rule, for which multiple declarations
+      // make sense (e.g. multiple `@font-type`s)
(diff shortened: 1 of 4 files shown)


artf:
Big thanks for the help Tom 👍

PR #1050, opened 2018-04-16
Update store after changes to css

dsgh:
Fixes https://github.com/artf/grapesjs/issues/1049

### DIFF
--- a/src/css_composer/index.js
+++ b/src/css_composer/index.js
@@ -96,17 +96,18 @@ module.exports = () => {
       const rules = this.getAll();
       em.stopListening(rules, ev, this.handleChange);
       em.listenTo(rules, ev, this.handleChange);
-      rules.each(rule => this.handleChange(rule));
+      rules.each(rule => this.handleChange(rule, { avoidStore: 1 }));
     },
 
     /**
      * Handle rule changes
      * @private
      */
-    handleChange(model) {
+    handleChange(model, opts = {}) {
       const ev = 'change:style';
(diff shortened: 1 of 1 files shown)


artf:
Great, thanks for the help Duarte

dsgh:
@artf Thanks for the merge!

I take it that one of the goals of grapesjs is to provide the ability to do code editing - either through core or through a plugin. I've created a repo with an idea of how that could work: https://github.com/portablemind/grapesjs-code-editor

I'd like to know what you think of it, and if you have something else in mind? I'm keen to make this happen in a way that's consistent with your vision.

artf:
Really cool @dsgh I like it. Honestly, my main goal is to make GrapesJS more usable for people who don't know how to code, but at least your plugin shows a more advanced usage. I would only suggest using the [standard way of initializing plugins](https://github.com/artf/grapesjs-plugin-boilerplate/blob/master/src/index.js#L1-L5) :)

dsgh:
@artf The thing is that this "plugin" doesn't define any blocks or components - it instead just exports a `codeCommand` object and a `panels` object, to be used in the grapesjs init block. In fact this is not a plugin at all, given that the plugin system has no way of 'exporting' commands or panels.

The plugin system would either have to change to support this usage, or this 'plugin' would be rolled into core - potentially in a different form - if there's interest.

artf:
Well yeah it makes sense, the plugin would just allow adding commands and panels transparently

Issue #1053, opened 2018-04-18
[Question]


FrciSmrci:
Hey @artf,

is there any flag available to leave the unused styles?  Referencing to you provided answer [here](https://github.com/artf/grapesjs/issues/843).

Thanks in advance and have a great day.

arthuralmeidap:
Hi @artf ,

  Added a basic support for this. Take a look on the #1054  and let me know what do you think

arthuralmeidap:
hey @FrciSmrci ,

  Just to let you know... I just added support for this for grapesjs. I created a new PR and I'm waiting for Artur's review and approval.

FrciSmrci:
Hey @arthuralmeidap , thanks for the pull request and letting me know. :) Have a great day.

artf:
Thanks @arthuralmeidap for the help, in the next release, this option will be available via `config.dumpUnusedStyles`

FrciSmrci:
@arthuralmeidap @artf Thank you both for the quick reply. :) @artf When could I expect the next release?

FrciSmrci:
@arthuralmeidap @artf I built the dev branch and tested the implementation in my project, for all I can tell when the flag dumpUnusedStyles is set to true it leaves them and when it's set to false it removes them. According to the dumpUnusedStyles naming I would say that this is an mistake? Or am I missing something? 😐

arthuralmeidap:
hey @FrciSmrci , how are you?

  I'm not a native english speaker, maybe the naming is confusing. For me makes sense. What do you think?
  
  For me, when I say: " dump unused styles" , it means export the used + unused styles from the editor.
Is this makes sense for you?

  The behaviour is correct. For now when you set as true it should leaves them and you set to  false it won't dump the unused styles.

FrciSmrci:
Hey @arthuralmeidap , I'm great thanks for asking. And you? :) 

Ohh now I understand what you meant, something like keepUnusedStyles/preserveUnusedStyles. Hmm I read dumpUnusedStyles as dispose them, because dump verb according to the dictionary means dispose of (rubbish, waste, or unwanted material), typically in a careless or hurried way. 

Otherwise I'm really grateful for this addition, thank you! You spared me a lot of time.

arthuralmeidap:
okay my friends, doing now! Thanks for the feedback and the english lesson!

PR #1054, opened 2018-04-18
WIP: Dump unused styles

arthuralmeidap:
Hi @artf ,

  As you may know, after import a html within grapesjs I need to dump those styles even if they aren't being used by any component within the canvas.

  I have made some basic changes to be able to dump those unused classes when I call `editor.getCss()` .

  This PR is still WIP because I would love to hear some feedback from you, to see what you think about it and if it has chances to be merged back on dev branch. 

   If you agree on this, I will improve the tests and create a parameter to be able to configure that when calling `grapesjs.init` also.

  Let me know what do you think!

  This is a PR #1053 related to that

### DIFF
--- a/src/code_manager/model/CssGenerator.js
+++ b/src/code_manager/model/CssGenerator.js
@@ -59,7 +59,7 @@ module.exports = require('backbone').Model.extend({
           return;
         }
 
-        code += this.buildFromRule(rule, dump);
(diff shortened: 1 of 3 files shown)


artf:
LGTM

arthuralmeidap:
hey @artf , it's done!

Take a look and thanks for the quick feedback!

PR #1130, opened 2018-05-24
Fix cloning edit image - pencil button while cloning an image

JuanTincho:
Fixing bug: https://github.com/artf/grapesjs/issues/1129.

**Add condition to check if the edit image button already exists**


artf:
Great, thanks Juan 👍

PR #1185, opened 2018-06-08
Add option to specify a rootNode for layerManager

### LINKED ISSUE #1096: [FEATURE REQUEST] Support rendering layouts which are not part of the editable content
Basically my use case is that I want to show a whole page - with header, menu(s) and footer - in the editor canvas, but only allow the user to edit the actual content of it. The reasoning behind this is so that the users can see the page as close as it will be to the "end product".

I've been able to accomplish something like this by rendering the full page into the wrapper, using the `fromElement` option, and taking care to set `highlightable`, `badgable` and all the other such attributes to `false` on the elements belonging to the layout, but still this is not ideal as:

1. The root node in the "Layer Manager" is still the body, which is then un-expandable as the elements directly below it are marked as non-layerable. What I'm looking for here is that the root of the Layer Manager be the "content" element.

2. When viewing/exporting code, the code for the whole page is shown. What I'm looking for is to have only the code related to the "content" element appear.

@artf Is this something that makes sense as a feature? I'm ready to start working on a PR for it. Should this be implemented as an option that specifies a css selector for the element that should be the "actual editable portion"? If so, what should the option be called? `rootNode`, `subcontainer`, `wrapper`, `subwrapper`, `contentElement`, `contentWrapper`? Or, if not a config option, then there could be a documented default behaviour. For example: if one of the elements of the initial html content (when using `fromElement`) has an id of, say, `gjs-root-node`, then the behaviour described above is applied accordingly. Thoughts?

dsgh:
Fixes #1096 

There's probably room for improvement of this PR on the `recursiveFind` method. I've implemented this as I didn't find the proper way to do it in Backbone, but I assume there must be a better way? Any pointers on this are welcome.

How to use this:

```html
<div id="gjs">
  <div class="menu-container">
    <div class="side-menu"></div>
    <div id="my-root"><!-- Users will append content here --></div>
  </div>
</div>
```

```javascript
  container: '#gjs',
  layerManager: { rootNodeId: 'my-root' },
  // ...
})
```

### DIFF
--- a/src/navigator/config/config.js
+++ b/src/navigator/config/config.js
@@ -14,6 +14,9 @@ module.exports = {
   // Hide textnodes
   hideTextnode: 1,
 
+  // Optionally specify the id of the node to show as root
+  rootNodeId: null,
+
   // Indicates if the wrapper is visible in layers
   showWrapper: 1
 };
--- a/src/navigator/index.js
+++ b/src/navigator/index.js
@@ -23,14 +23,35 @@ module.exports = () => {
     onLoad() {
       const collection = em.get('DomComponents').getComponents();
       const parent = collection.parent;
+      let rootNode;
+
+      function recursiveFind(object) {
+        if (
+          object.attributes &&
+          object.attributes.attributes.id === config.rootNodeId
+        ) {
+          return object;
+        } else if (object.get('components')) {
+          return object
+            .get('components')
+            .models.map(recursiveFind)
+            .filter(function(n) {
+              return n != null;
+            })[0];
+        }
+      }
+
       const options = {
         level: 0,
         config,
         opened: config.opened || {}
       };
 
-      // Show wrapper if requested
-      if (config.showWrapper && parent) {
+      if (config.rootNodeId) rootNode = recursiveFind(parent);
+      if (config.rootNodeId && rootNode) {
+        View = ItemView;
+        options.model = rootNode;
+      } else if (config.showWrapper && parent) {
         View = ItemView;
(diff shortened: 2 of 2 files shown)


artf:
Thanks for the effort Duarte, but I think I've figured out a better way to accomplish the same task (inspired by your idea btw) so I close this PR. I've created a new branch where I refactored few old ideas and implemented the new one: https://github.com/artf/grapesjs/blob/dynamic-layer-root/src/navigator/index.js
I plan to merge it for the next release

Basically, for the initialization, a query string is enough as I moved all the logic in `LayerManager.postRender()`
```js
  container: '#gjs',
  layerManager: { root: '#my-root' },
  // ...
})
```
Besides this, you can also change the root after
```js
editor.LayerManager.setRoot('#other-root');
// or via model
editor.LayerManager.setRoot(editor.getSelected());
```

chiqui3d:
@artf I need your advice please, I've been reviewing this Issues and the one it refers to, but I can't find a way to make it easier and make it work.

I'll give you my example:

In identifier #gjs, I load all the content, including the header, footer and sidebar, but I just need to change the content that is under the myeditable ID, I have tried this, but I can still select and edit things from the header, well everything in general.

I don't know if I explained myself, I hope you can help me solve this problem, as this is the best way to edit a page which has its under a CMS.

artf:
@chiqui3d You should basically setup **header**, **footer** and **sidebar** as placeholders
https://jsfiddle.net/5vcc3en6/

PR #1271, opened 2018-07-10
Added onIframeLoaded option

Patrick-Spiegel:
Adds a possibility to make use of the IFrame loaded event. 

#### Vuejs can be used that way in the iframe
What could be made with that event is e.g. using the `renderChildren` from [this](https://jsfiddle.net/PatrickSpiegel/L9se31bx/1/) fiddle in the load function.

### DIFF
--- a/src/canvas/config/config.js
+++ b/src/canvas/config/config.js
@@ -39,7 +39,5 @@ module.exports = {
    *  // do something
    * }
    */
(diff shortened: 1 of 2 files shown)


Patrick-Spiegel:
Sorry for disturbing you. 
  `editor.on("loaded", function () {
    
  });`
That is sufficient for me.

PR #1276, opened 2018-07-11
Fix onHover

dsgh:
There's a reference to a `comp` variable which doesn't exist. I believe this PR fixes it.

Solves https://github.com/artf/grapesjs/issues/1275


artf:
Thanks Duarte 👍

PR #1301, opened 2018-07-20
Allow plugins to be defined as functions

loganvolkers:
As a developer that is using a lot of custom GrapesJS plugins, I would like to import plugins and use them directly without having to define them in the global plugin registry.

This allows my code to be more modular without having to worry about another plugin using the same string namespace that my plugin is trying to use.

Benefits:
- Avoids the possibility of namespace conflicts
- Lowers the bar for people to develop their first plugin
- Allows for vanilla javascript plugin composition and higher order plugins
- Plugins don't need to import and/or bundle Grapes [e.g. nicer boilerplate](https://github.com/artf/grapesjs-plugin-boilerplate/blob/master/src/index.js#L5)

## Using imports
```
  import grapesPluginFn from "./awesomePlugin";

      container : '#gjs',
      plugins: ['my-plugin-name', grapesPluginFn ]
  });
```

## Using local plugins
```
  function customizeMyEditor(editor,options){
     editor.BlockManager.add('test-block', {
        ...
     });
  }

      container : '#gjs',
      plugins: ['my-plugin-name', customizeMyEditor]
  });
```

## Lambda plugins
```
      container : '#gjs',
      plugins: ['my-plugin-name', (editor) => console.log("Pimping the editor before it renders", editor)]
  });
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -58,6 +58,8 @@ module.exports = (() => {
 
         if (plugin) {
           plugin(editor, config.pluginsOpts[pluginId] || {});
+        } else if (typeof pluginId === 'function') {
+          pluginId(editor, config.pluginsOpts[pluginId] || {});
         } else {
           console.warn(`Plugin ${pluginId} not found`);
         }


src/index.js
```diff
@@ -58,6 +58,8 @@ module.exports = (() => {
 
         if (plugin) {
           plugin(editor, config.pluginsOpts[pluginId] || {});
+        } else if (typeof pluginId === 'function') {
+          pluginId(editor, config.pluginsOpts[pluginId] || {});
```
artf: In this case, you can't pass options, so `config.pluginsOpts[pluginId]` doesn't make sense.
You could pass them only in this way
```js
plugins: [
	myPluginNoOptions,
	e => myPluginWithOptions(e, { ... }),
]
```

loganvolkers: I was on the fence about this too.

- On one hand you can use objects as keys (see: https://jsfiddle.net/n0ou7x84/)
- On the other hand, this is a bit of an obscure syntax for most people

I left it in because I didn't imagine that it would cause much harm, the options may be null or ignored in many cases.

In terms of docs, though, I think the option-passing approach you explained is the more user-friendly approach and should be included as the recommended way of passing in plugin options.

loganvolkers:
Thanks for the feedback, I'll see about adding some tests and docs soon.

loganvolkers:
FYI - I might have gone too far with the docs. I think that this new functional approach to plugins is much nicer that the old namespace approach, and I'd like to see all plugins migrated this way.

That is in part because in our project we have problems with this:
```js
export default grapesjs.plugins.add('my-plugin-name', (editor, options) => {
  console.log(options);
  //{ customField: 'customValue' }
})
```

Because we use a bundler and lots of tests, it's a pain to always force `grapesjs` into global scope, causing workaround like this:

```js
import grapesjs from 'grapesjs'
windows.grapesjs  = grapesjs

function onLoad(){
   require("grapesjs-preset-newsletter");
}
```

That sounds like a bigger project, but I figured it would be a good idea to share.

loganvolkers:
@artf Can you have a look at this CI error? Tests are working locally.

```
    SecurityError: localStorage is not available for opaque origins
```

^ that's the error that I'm getting, and it sounds unrelated to the feature I added

loganvolkers:
> ... still not sure about using it as a recommended way, especially when you use plugins directly in the browser
> `plugins: [ ?? ] // would require some kind of global`

To be clear, your plugin already provides a global export UMD module by default.

## Current state

 - UMD namespace: `window["grapesjs-preset-newsletter"]` (current UMD naming - [source](https://github.com/artf/grapesjs-preset-newsletter/blob/master/dist/grapesjs-preset-newsletter.min.js))
 - NPM namespace: `npm install grapesjs-preset-newsletter`
 - Grapes Plugin namespace: `plugins: ['gjs-preset-newsletter']`

Plugins can use either `plugins: [window["grapesjs-preset-newsletter"]]` or `plugins: ["gjs-preset-newsletter"]`

## Proposed state

If you embraced plugins as functions, you would remove 1 of your 3 namespaces.

```
    plugins: [window.grapesjsPresetNewsletter]
```

Ideally your plugins and boilerplate would use a more sane UMD export.

 - UMD namespace: ~~`window["grapesjs-preset-newsletter"]`~~ `window.grapesjsPresetNewsletter` (better UMD naming)
 - NPM namespace: `npm install grapesjs-preset-newsletter`
 - ~~Grapes Plugin namespace: `plugins: ['gjs-preset-newsletter']`~~

loganvolkers:
> `const myPlugin = editor => { ... };`
> `window.myPlugin = myPlugin;`
> `export default myPlugin;`
> correct?

Not quite correct. Webpack will turn `export default myPlugin` into `window.myPlugin = myPlugin;`. 

There should be no need to ever manually add library to `window` in modern JS anymore, that should always be handled by webpack, or rollup, or parcel.

> BTW, to be honest, I didn't even remember the global definition made by UMD wrapper, so in that case, when you ask for plugins: ["gjs-preset-newsletter"] instead of just looking for it inside grapesjs.plugins we can also check inside window... what do you think?

So `plugins: ["gjs-preset-newsletter"]` would be equivalent to `plugins: window["gjs-preset-newsletter"]`?

That seems reasonable to me.

artf:
> So plugins: ["gjs-preset-newsletter"] would be equivalent to plugins: window["gjs-preset-newsletter"]?

I'd say more like this `plugins: [ window["gjs-preset-newsletter"] ]`, the `plugins` options should always be an array

In that case, for the next release, I'll add the check inside the `window` and, later, update the [grapesjs-plugin-boilerplate](https://github.com/artf/grapesjs-plugin-boilerplate) to clean it from `peerDependencies`

PR #1359, opened 2018-08-13
Prevent error when there's no activeElement during undo/redo

nojacko:
Our errors logs picked up a lot of "Cannot read property 'activeElement' of null" errors so I put in a check.

I did spot the comment above the function:
```
    /**
     * Detects if some input is focused (input elements, text components, etc.)
     * Used internally, for example, to avoid undo/redo in text editing mode
     * @return {Boolean}
     */
```
The function only seems to return true is tag isn't BODY which doesn't match the comments. Is this correct?

### DIFF
--- a/src/canvas/index.js
+++ b/src/canvas/index.js
@@ -363,7 +363,11 @@ module.exports = () => {
      * @return {Boolean}
      */
     isInputFocused() {
-      return this.getFrameEl().contentDocument.activeElement.tagName !== 'BODY';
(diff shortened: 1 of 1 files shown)


artf:
Yes James, that was actually an intentional behaviour, if activeElement was different from `Body`, some editable component, inside the canvas, was selected. Actually I had other issues with this method and now I prefer to rely on `em.isEditing`, probably I'll replace it in Undo Manager in the next release and will see what to do with this method, for now I accept you PR as it's a good check in any way

PR #1375, opened 2018-08-22
Don't display drop placeholder when no target

### LINKED ISSUE #1361: [Bug]: The drag placeholder should not be visible when there's no target
I'm using the latest release.

Steps to reproduce:
- Configure grapesjs in such a way that the wrapper is not droppable.
- I have created a fiddle to demonstrate this: https://jsfiddle.net/szLp8h4n/300/ (the idea here is that only the white center area is droppable)

Expected behaviour:
- The green drag placeholder shouldn't be visible when not dragging over an acceptable target

Screenshot:
green placeholder
(note: this is paused in debugger to be able to catch the green drag placeholder. Bottom line is that it shouldn't be there in this case :)

I will submit a Pull Request to fix this.

dsgh:
Fixes #1361

### DIFF
--- a/src/utils/Sorter.js
+++ b/src/utils/Sorter.js
@@ -385,6 +385,7 @@ module.exports = Backbone.View.extend({
     const target = this.target;
     const targetModel = this.getTargetModel(target);
     this.selectTargetModel(targetModel);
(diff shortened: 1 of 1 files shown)


artf:
👍

PR #1392, opened 2018-08-29
Update Traits.md

cmoutafidis:
Changed the method name to "onChange" so it is clear that the name of the method comes from the events key value object.


artf:
`onChange` is triggered when you `keyup` from the input, `onValueChange` instead when the `value` of the model is changed. The example is just showing the existence of that method :)

PR #1415, opened 2018-09-11
Fixed height/width calculations when height/width not numeric values.…

nojacko:
… Allows for the auto height/width to be preserved.

Problem I had: When I added images to Grapes, I'd set `width: <image's width>; max-width: 100%; height: auto;` so the image works well with responsive designs.

1. When the attempting to resize by dragging, some values got set to `NaNpx`. This solves that.
1. It also gives to option to `keepAutoHeight` and (form completeness, don't think i've see `width: auto` used before) `keepAutoWidth` which prevents height or width changing if the values were 'auto'. Defaults to off to preserve current behaviour.

### DIFF
--- a/src/commands/view/SelectComponent.js
+++ b/src/commands/view/SelectComponent.js
@@ -370,14 +370,30 @@ module.exports = {
         // Here the resizer is updated with the current element height and width
         onStart(e, opts = {}) {
           const { el, config, resizer } = opts;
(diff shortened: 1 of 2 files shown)


artf:
Very cool, I see you used `widthAuto` in `SelectComponent.js` instead of `autoWidth` but it's ok, I'm gonna fix it :)
Thanks @nojacko 👍

nojacko:
Nice spot. Thanks.

PR #1417, opened 2018-09-11
Only change onbeforeunload when that option is set

dsgh:
Fixes https://github.com/artf/grapesjs/issues/1416


artf:
Cool, thanks @dsgh 👍

PR #1430, opened 2018-09-17
S3 compatible asset manager

tar672:
This PR are the changes I had to make to support using an S3 bucket as the host server for assets.

- When using [S3 POST](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html) CORS means that permission is denied when including auth cookies so credentials must be set false.
- The endpoint only allows 1 file at a time and insists on the name of the file being `filename`. It cannot be `filename[]` so I added a multiUpload flag.
- The endpoint also insists on `key` being the first value in the the form so I moved the the files to after the params in body formdata.
- The endpoint returns an empty document which unsurprisingly is not valid JSON so I stuck that parse in a try catch.

Happy to provide an example/some docs for how to hook it up to S3 although its not trivial.

Thanks so much for a great project, couldn't believe how perfect it was for my needs when I found it!

### DIFF
--- a/docs/modules/Assets.md
+++ b/docs/modules/Assets.md
@@ -66,6 +66,10 @@ Below the list of currently available options
   // The credentials setting for the upload request, eg. 'include', 'omit'
   credentials: 'include',
 
+  // Allow uploading multiple files per request.
+  // If disabled filename will not have '[]' appended
+  multiUpload: true,
+
   // If true, tries to add automatically uploaded assets.
   // To make it work the server should respond with a JSON containing assets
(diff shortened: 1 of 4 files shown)


artf:
Hi Tom, thanks for your great PR, I'm pretty sure such config might also be useful for someone else 👍

Controlhaus:
@tar672 any chance you could provide an example?

ThunderDev1:
Hi @tar672 thanks for the PR, would also be interested in some sample code for hooking up to S3 :)

PR #1437, opened 2018-09-18
Remove Links in RTE

tabaliov:
This is a quick fix that allows removing links in RTE. Tested in my project worked well and decided to share the fix. Opened to any suggestions for a better implementation as i learned about the Seleciton API on the go.


artf:
Great 👍, thanks @tabaliov I like this improvement

PR #1482, opened 2018-10-06
Add event component:destroy to ComponentDelete

david-polak:
It is impossible to distinguish between remove() called on component when moving and when removing.


artf:
👍

PR #1518, opened 2018-10-15
Update Traits.md

jereddanielson:
Previous example does not work. There is no `change:model-prop-name` event. The proper event is `change:attributes:model-prop-name`.


artf:
Thanks for the PR @jereddanielson but it's actually correct.
By default, traits are appending to attributes so if you need to listen their changes `change:attributes:*` is the correct event, but in case you use `changeProp: 1,` in you trait definition, `change:*` is the one to use

PR #1551, opened 2018-10-30
Fullscreen command with a target

mcsekar12:
https://github.com/artf/grapesjs/issues/1526 
As suggested in https://github.com/artf/grapesjs/issues/1526#issuecomment-431193245

`// from
command: 'fullscreen',
//  #to
command: editor => editor.runCommand('fullscreen', { target: '.ogaeditor' }),
// The target might be a selector string or any HTMLElement`


artf:
You've only forget to remove `debugger;` but for the rest is ok, thanks @mcsekar12

PR #1552, opened 2018-10-30
Make grammar improvements

jacobherrington:
There is still more room for improvement, but making some more progress.


artf:
Thanks Jacob ❤️

Issue #1602, opened 2018-11-22
[QUESTIONS] - Panel visibility


arthuralmeidap:
Hey @artf ,

  How have you been?
  Quick question: a Panel has a property called `visible`. I'm trying to set it to false but nothing happens.
  I'm getting the panel after the  GrapesJS is rendered like `editor.Panels.getPanel('myId').set('visible', false)`;
  I was digging into the source code and I didn't find any use to the `visible` property for the Panel model. 
  Am I missing something?
  Thanks in advance

arthuralmeidap:
Oh, I didn't say but I want to hide/show the panel according to some condition.

I know how  I could get the panel html element itself and change the `display` but I was wondering about the `visible` property

artf:
> Toggle the style display property between none and block?

Yeah I think it's enough, but in case it's `true` just don't set any `display` (eg. if someone had customized the panel with `flex` that would break its layout)

arthuralmeidap:
created MR https://github.com/artf/grapesjs/pull/1603

PR #1603, opened 2018-11-24
add the ability to hide/show a panel by the  panel's visible property

arthuralmeidap:
Related issue #1602


artf:
Thanks Arthur 👍

PR #1644, opened 2018-12-10
Add component:clone propagation

david-polak:
Event `component:clone` should be propagated to it's children. Currently there is no way to distinguish between clone and create.

I would recommend against removing the `component:clone` event at [PasteComponent.js](https://github.com/artf/grapesjs/blob/dev/src/commands/view/PasteComponent.js) as that is triggered after the view is also initialized and might break previous functionality dependent on this behaviour.


artf:
Yeah it makes much more sense, probably I'm just gonna update the old one as `component:paste`

PR #1656, opened 2018-12-12
Add default units vh and vw

david-polak:
Add `vh` to vertical properties and `vw` to horizontal properties, they are essential for nested one-page design.


artf:
Thanks David 👍

Issue #1615, opened 2018-11-28
[Feature] support edit credentials request attribute on Remote Storage


jmtt89:
The credentials options its forced to be 'include' on Remote Storage. 

https://github.com/artf/grapesjs/blob/master/src/storage_manager/model/RemoteStorage.js

```
    fetchOptions = {
      method: opts.method || 'post',
      credentials: 'include',
      headers
    };
````

arthuralmeidap:
Created MR #1705

PR #1705, opened 2019-01-03
added support to configure credentials option for Remote Storage

arthuralmeidap:
Hey @artf ,

  This is related to issue #1615 .
  How could I improve the docs for this?


artf:
Thanks for the help @arthuralmeidap great and clean as always 👍 

> How could I improve the docs for this?

I think it's all good as we point to the configuration source file here: https://grapesjs.com/docs/modules/Storage.html

PR #1710, opened 2019-01-04
Add support for display:flex

david-polak:
This implements complete support for flex.

- Added new Sector **Flex** 
To serve as a Sector for properties of both the `flex` container and it's children.

- Added dynamic styleable dependencies
  - Added a new configuration option to PropertyFactory

```JavaScript
  obj.requires = { display: ['flex'] };
```
This is needed for hiding the Flex sector options when display of the selected object is not `display: 'flex'`.

- Added dynamic styleable dependencies for parent
  - Added a new configuration option to PropertyFactory 

```JavaScript
  obj.requiresParent = { display: ['flex'] };
```

This is needed because we need to detect if the current selected object is a direct descendant of a `display: 'flex'` object.

While the adding the styleable dependencies for the current object is straight forward, adding styleable dependencies for its parent is not trivial, I was forced to use 
```JavaScript 
const styles = window.getComputedStyle(trg.view.$el[0].parentNode);
```
which is not ideal, but it works without major intrusions into the existing model.


david-polak:
Right, will add tests.

src/style_manager/view/PropertyView.js
```diff
@@ -427,6 +439,37 @@ module.exports = Backbone.View.extend({
...
     }
 
+    // Check if the property is available based on other property's values
+    if (sectors && requires) {
+      const properties = Object.keys(requires);
+      sectors.each(sector => {
+        sector.get('properties').each(model => {
+          if (properties.includes(model.id)) {
+            const values = requires[model.id];
+            stylable = stylable && values.includes(model.get('value'));
```
artf: `Array.includes` is not available in IE, so use `import { includes } from 'underscore'` instead

src/style_manager/view/PropertyView.js
```diff
@@ -427,6 +439,37 @@ module.exports = Backbone.View.extend({
...
+    // Check if the property is available based on parent's property values
+    if (requiresParent) {
+      if (
+        trg.view &&
+        trg.view.$el &&
+        trg.view.$el[0] &&
+        trg.view.$el[0].parentNode
+      ) {
+        const styles = window.getComputedStyle(trg.view.$el[0].parentNode);
+        Object.entries(requiresParent).forEach(([property, values]) => {
```
artf: `Object.entries` is not available in IE, use `_.each(requiresParent, (values, property) => {...})`

src/style_manager/model/PropertyFactory.js
```diff
@@ -270,6 +300,29 @@ module.exports = () => ({
...
+          obj.requires = { display: ['flex'] };
+          break;
+        case 'order':
+        case 'flex-grow':
+        case 'flex-shrink':
+        case 'align-self':
+          obj.requiresParent = { display: ['flex'] };
+          break;
+      }
+
```
artf: It would be great if you add those properties to `defaults` (in style_manager/model/Property.js) and write a tiny comment about the usage, otherwise, we risk to forget about them

david-polak:
Weird, I am currently investigating a bug, whereby clicking on a flex item that has been loaded from html the view isn't available.
screenshot_20190107_172710

It only happens under specific circumstances, I am not yet sure exactly what those are. Could it be Block related?

david-polak:
`this.em.getSelected()` seems to be correct in all cases I could test.

artf:
> It only happens under specific circumstances, I am not yet sure exactly what those are. Could it be Block related?

mmm the view is not available immediately on the model but only after the mounting, so I'm thinking if, somehow, you're reaching that part of code before the view is created (eg. maybe auto-selecting the component on canvas drop 🤔)

ps. I'd also suggest making that check easier to read:
```js
const parentComp = trg.parent();
const parentEl = parentComp && parentComp.getEl();
if (parentEl) {
	const style = ....
}
```

david-polak:
I think that's about it, anything else to be done for flex support?

david-polak:
> > Weird, I am currently investigating a bug, whereby clicking on a flex item that has been loaded from html the view isn't available.
> 
> @david-polak so I didn't get if you face the issue after the mounting

I changed the selected element detection to `this.em.getSelected()` and now it works.

artf:
Ah sorry, I didn't get it was already solved 😬 
Just fix that conflict and I'm gonna merge it :)

artf:
Thanks @david-polak great work

Issue #1728, opened 2019-01-14
[FEATURE REQUESTS] -  block's toolbar at bottom option.


arthuralmeidap:
hi @artf 

  How are you?
  I'm facing a small issue on GrapesJS and I would like to hear your thoughts about the best approach here.
  My issue is: When I have a small block, like a Link, placed at the top left corner of the Canvas, the toolbar totally covers the block and I can't see it. I would have to unselect that block to see it again.
  What do you think as the best solution to this problem? I was thinking about to create a new option for GrapesJS to let the user select to display the toolbar at the bottom or the top of the block. This fixes my bug but introduces another, that's when the block is at the bottom of the canvas I would have to display the toolbar at the top this time.
    What do you think? Is there any chance this new option be added to GrapesJS code? Do you have  a better approach? Please let me know!

artf:
Hi @arthuralmeidap I think the #1434 talks about the same problem.
Honestly, I've already solved, basically the same issue, with the RTE toolbar, so I guess, if you have time, you could copy/share the same functionality for the Component's toolbar.

I'll place below a few hints about the current implementation in RTE:
Currently, if you activate the RTE, its toolbar will be placed at the top. Once the top canvas edge is reached, the toolbar is placed below the component.
Reproducible on the [demo](https://grapesjs.com/demo.html)
Implemented here: https://github.com/artf/grapesjs/blob/8af686cb7fa66863c04b786ed1aa704e5406252e/src/rich_text_editor/index.js#L234-L253

As you might see I pass also `rteToolbarPosUpdate` event to the `getTargetToElementDim` method, which allows to customize the position extarnally (example usage in the CKEditor plugin: https://github.com/artf/grapesjs-plugin-ckeditor/blob/7c227de0b187da8cb7b1e22b3da34837659977d1/src/index.js#L110-L132)

The only problem with this implementetion is when you have to deal with big component (eg. section), not that common in text components. For instance, when you reach the top canvas edge but the bottom of the component it's still below the canvas. So, in that case, it might be good to leave the current positioning: https://github.com/artf/grapesjs/blob/8af686cb7fa66863c04b786ed1aa704e5406252e/src/commands/view/SelectComponent.js#L529-L544

arthuralmeidap:
@artf I created the PR #1730 for this. Take a look when possible! Thanks for your help as always!

PR #1730, opened 2019-01-17
Fixed display for the component's toolbar to not cover component's content

arthuralmeidap:
Related to issue #1728


artf:
Thanks Arthur, well done 👍

PR #1736, opened 2019-01-21
limit width resize to Canvas width

arthuralmeidap:


artf:
Great, thanks @arthuralmeidap

PR #1755, opened 2019-01-27
try undo

thinkanymore:
Try to solve undo,which has script doesn't work


artf:
Great, I see it works now 👍

PR #1773, opened 2019-02-05
Add embedAsBase64 to AssetManager config

david-polak:
This should definitely be in the config file.

I argue that it should be defaulted to true, because right now the upload box in the default configuration does not actually work.


src/asset_manager/config/config.js
```diff
@@ -63,6 +63,9 @@ module.exports = {
   // }
   uploadFile: '',
 
+  // In the absence of uploadFile function assets will be embedded as Base64
+  embedAsBase64: 1,
```
artf: Somehow I've missed to add it to the config file, thanks.
BTW we need it false as default. For a correct setup you just have to provide the upload endpoint but with this option as true the uploading will not work (it will be even a breaking change), so we'll get a lot of complaints that uploading is not working

src/asset_manager/view/FileUploader.js
```diff
@@ -32,7 +32,7 @@ module.exports = Backbone.View.extend(
 
       if (uploadFile) {
         this.uploadFile = uploadFile.bind(this);
-      } else if (c.embedAsBase64) {
+      } else if (!c.upload && c.embedAsBase64) {
```
david-polak: @artf perhaps this way it would be satisfactory. `embedAsBase64` gets attached as long as `uploadFile` or `upload` is not configured.

This way there wouldn't be a non-functioning drop area by default.

artf: Good idea, I like it 👍

PR #1790, opened 2019-02-13
Add category on traits

WebEtSolutions:
Hi !

We have added the possibility to make traits under categories like blocks. When you create a trait you can add a property like this

`
label: 'My trait',
type: 'text',
category: {
  id: 'general',
  label: 'General',
}
`

And the "My trait" trait will be make under the "General" category.

category propertie is optional and it's like the block category. (string or object)

### DIFF
--- a/src/selector_manager/view/ClassTagView.js
+++ b/src/selector_manager/view/ClassTagView.js
@@ -96,7 +96,8 @@ module.exports = require('backbone').View.extend({
     const { em, model } = this;
     const sel = em && em.getSelected();
     // Prevent weird erros on remove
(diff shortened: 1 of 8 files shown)


artf:
Sorry guys, I really appreciate your efforts, but I can't accept this PR. 
This is just a big copy paste from the BlockManager module, even with some logic IMHO that doesn't make sense to have (like private and public collections, it's ok for Blocks but not for Traits).
First of all, the correct approach for any PR (especially big like your) on OSS projects it should be to ask the team (by opening an issue) if this is something we need to add to the core. Then probably we would tell you ok and we would share some guidelines about the implementation (like creating an abstract module for categories and refactoring the BlockManager code).

Issue #1857, opened 2019-03-05
Protected class not used


WebEtSolutions:
Hi !

On your ClassTagView, you don't check the protected attribute before remove tag.

PR #1858, opened 2019-03-05
Check protected attribute before remove a tag

WebEtSolutions:
#1857


artf:
Thanks @WebEtSolutions 👍

PR #1883, opened 2019-03-12
add a listener to fire onViewClose when modal is closed on interface

lucasschirm:
I think when the user closes the modal using any interface functional like, click on the exit sign or outside the modal should fire the modal:close trigger.


artf:
Thanks @lucasschirm you're totally right, and this why I've already added it 😁
https://github.com/artf/grapesjs/blob/01a3053270b8b68c61091b4eb3445e2450b061f6/src/modal_dialog/index.js#L70
It will be available in the next release

lucasschirm:
Thank you! Glad to help!

Att,
Lucas Schirm


Em qui, 21 de mar de 2019 às 15:50, Artur Arseniev <notifications@github.com>
escreveu:

Issue #1914, opened 2019-03-22
[Bug] Change default font size unit to EM instead of REM.


lyczos:
Now in GrapesJS editor, we have some of font sizes in `REM`. I think they should be in `EM` unit because if we want to embed editor in the existing page there may be some issues with proper font size - in case when the parent font size is set to other then default size (ex. `:root { font-size: 10px }`). 
https://i.imgur.com/Vzoa5y9.png

artf:
Yeah, I think you're right, would you like to create a PR?

lyczos:
Yep, sure, I will do it later :)

PR #1923, opened 2019-03-26
bugfix(1914) Change REM to EM

lyczos:
#1914 Changed all REM units to EM units.

### DIFF
--- a/src/styles/scss/_gjs_blocks.scss
+++ b/src/styles/scss/_gjs_blocks.scss
@@ -34,7 +34,6 @@
   box-sizing: border-box;
   min-height: 90px;
   cursor: all-scroll;
-  font-size: 11px;
   font-weight: lighter;
   text-align: center;
   display: flex;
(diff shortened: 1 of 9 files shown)


lyczos:
@artf any info updates on this? :)

artf:
Hi @lyczos sorry for the late response. BTW would you like to check how to solve the issue in the block panel?
Schermata 2019-04-13 alle 18 22 32
I mean, I know what is happening but I'm not really sure how to solve it.
For example:
- The Text block is the standard behavior and is ok
- Link and Video have `fa` custom class on top (from the block definition I'm able to pass custom classes for the ) which has another font-size (but take always in consideration that the user can pass its own classes with other sizes)
- Form and input blocks are with a nested `gjs-block-label` class 😬

lyczos:
@artf sure, I will look into it.

larsdecker:
Any news about this topic?

For projects that will include GrapesJS as a Library, the changes will be good.

PR #1944, opened 2019-04-04
Fixed a simple spelling error. Mounted was spelled "monted".

nehoffman:
Fixed a simple spelling error. Mounted was spelled "monted". If you're wondering what the extra changed lines in dist/grapes.js, so am I. Looks like nano arbitrarily made a white space change of some sort. Sorry about that.


artf:
Hi @nehoffman, in this case, all you had to do is just to update the `src/editor/index.js` file, all other files are just auto-generated. I'm gonna close this PR just to keep things clean, but thank you very much for the fix.

PR #1957, opened 2019-04-09
Calrify docs

benvmatheson:


artf:
Hi Ben, thanks for the PR I'm gonna merge it but I'll fix the source as all the `/api` files are generated from their relative modules, so in this case the file to edit is [`selector_manager/index.js`](https://github.com/artf/grapesjs/blob/dev/src/selector_manager/index.js)

PR #2077, opened 2019-06-13
Missplaced comma fix

jcamejo:


artf:
@jcamejo great, thanks

BTW those API .md files are generated from the code, so you should actually change this line https://github.com/artf/grapesjs/blob/60c1755c57da02b57988fbe0e1a0321a5395728c/src/rich_text_editor/index.js#L141 
I'll merge it after this change 😬

jcamejo:
Sure!

PR #2108, opened 2019-07-01
Issue #2002 Resizer padding fix

### LINKED ISSUE #2002: Images won't resize to full width
When an image is resized with the mouse, it won't stretch all the way to the edge of the canvas, leaving a white space on the right side. If adjusting Dimensions --> Width, you can bring it all the way to the edge of the canvas.

image

If you drag enough elements onto the canvas for the scrollbar to appear, the scrollbar fills the white space that the image wouldn't fill. It appears the dragging is just taking the scroll bar into consideration and won't fill the scroll bar's area, even if the scroll bar doesn't exist yet.

image

https://jsfiddle.net/benvmatheson/2r3jdfpa/6/

giorgiosjames:
Fixes #2002 

Pull #1736 added Resizer limit for canvas width.
10 pixels of padding was hard-coded in, presumably to stop Resizer handles from being moved partially outside the canvas. 
However, this padding removes the ability for a component to be resized to the width of the canvas, leaving a 10px gap, as per Issue 2002.

This pull adds the padding value to Resizer config, so the padding can be removed if need be, opt-out.

### DIFF
--- a/src/commands/view/SelectComponent.js
+++ b/src/commands/view/SelectComponent.js
@@ -453,12 +453,11 @@ module.exports = {
           const style = modelToStyle.getStyle();
 
           if (!onlyHeight) {
-            const padding = 10;
             const frameOffset = canvas.getCanvasView().getFrameOffset();
             const width =
-              rect.w < frameOffset.width - padding
+              rect.w < frameOffset.width
                 ? rect.w
-                : frameOffset.width - padding;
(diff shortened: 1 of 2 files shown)


artf:
Explained here the reason https://github.com/artf/grapesjs/issues/2002#issuecomment-525058148
BTW thanks for the PR

PR #2161, opened 2019-07-23
Absolute mode: editor is not defined

martysweet:
In relation to https://github.com/artf/grapesjs/issues/2160

Change runCommand to reference the passed in ed

Let me know if you need anything else :)


artf:
Thanks @martysweet tiny thing but for sure would be hard to spot if you expose always the editor instance (like I do 😁)

PR #2197, opened 2019-08-16
fix url error

justinzzc:
fix url error with value which includes ( or )


justinzzc:
find some exception when url includes '

PR #2219, opened 2019-08-27
Compound selectors may no longer be extended

thehelvetian:
- Dart Sass 1.0.0
- Ruby Sass 4.0.0

See https://sass-lang.com/documentation/breaking-changes

### DIFF
--- a/src/styles/scss/_gjs_selectors.scss
+++ b/src/styles/scss/_gjs_selectors.scss
@@ -3,7 +3,7 @@
 }
 
 .#{$clm-prefix}select {
-  @extend .#{$sm-prefix}field.#{$sm-prefix}select;
+  @extend .#{$sm-prefix}field, .#{$sm-prefix}select;
 }
 
 .#{$clm-prefix}tags {


artf:
@thehelvetian can you explain was the issue this PR should fix? The change seems also to break the style (the extend will take other styles)

thehelvetian:
Hey @artf,

this happens when I `@import '~grapesjs/src/styles/scss/main';` into one of my projects and have webpack build the assets. The error comes from Dart Sass (sass@1.22.10).

Screen Shot 2019-09-02 at 11 38 46

The change in the PR does not break any styles (at least here). The intended styles are being applied correctly:

Screen Shot 2019-09-02 at 12 01 14

More info here:
- https://sass-lang.com/documentation/at-rules/extend#disallowed-selectors
- https://sass-lang.com/documentation/breaking-changes/extend-compound

HTH

artf:
> this happens when I @import '~grapesjs/src/styles/scss/main'; into one of my projects and have webpack build the assets. The error comes from Dart Sass (sass@1.22.10).

Ok now it's clear

> The change in the PR does not break any styles (at least here). The intended styles are being applied correctly:

The target is not the same (should be `.gjs-clm-select`) but seems like it's not used anymore, so I'll check it better later

PR #2224, opened 2019-08-28
migration to typescript

ayazhussein:
Hi @artf 
I like your project and I wanted to help to migrate the project to typescript so that it would be easier to use in developing other apps with types and also it offers easier documentation and IDE support.
Let me know if I should pursue this.


src/editor/config/config.ts
```diff
@@ -9,7 +204,7 @@ export default {
   style: '',
 
   // If true, will fetch HTML and CSS from selected container
-  fromElement: 0,
+  fromElement: false,
```
artf: Can we have multiple types here? Being a client-side javascript, use 1/0 as a boolean is preferable

artf:
BTW, isn't better declaring types in the [types repository](https://github.com/DefinitelyTyped/DefinitelyTyped)?

ayazhussein:
hmm, it might be but I was thinking that it would improve development and also I get to know the framework.

artf:
> hmm, it might be but I was thinking that it would improve development and also I get to know the framework.

well, it would be the same also by using types repository :)

src/editor/config/config.ts
mamal72: @artf Can I ask why? The boolean is more explicit here and it's meant to be used in binary cases like this.

src/index.ts
```diff
@@ -44,7 +45,7 @@ export default {
    *   style: '.hello{color: red}',
    * })
    */
-  init(config = {}) {
+  init(config:EditorConfig) {
```
mamal72: Missing space after `:`.

```suggestion
  init(config: EditorConfig) {
```

src/editor/config/config.ts
```diff
@@ -1,4 +1,199 @@
-export default {
+export interface EditorConfig {
+
+  /**
+   * The html target element.
+   */
+  container?: string;
+
+  plugins?: any[];
```
mamal72: I think we can have better typing here for plugins.

src/editor/config/config.ts
```diff
@@ -1,4 +1,199 @@
...
+  // 'translate' - Use translate CSS from transform property
+  // To get more about this feature read: https://github.com/artf/grapesjs/issues/1936
+  dragMode?: boolean;
+
+  // Dom element
+  el?: Element | string,
+
+  // Configurations for Undo Manager
+  //TODO: convert configs to ts
+  undoManager?: any,
```
mamal72: I think we can/should have some explicit types for this and the following keys. Having it as any does not provide much value for the users.

src/editor/config/config.ts
```diff
@@ -1,4 +1,199 @@
...
+
+  //Configurations for Asset Manager
```
mamal72: Missing space character after `//` in this line and a number of following lines.

src/editor/config/config.ts
artf: > @artf Can I ask why? The boolean is more explicit here and it's meant to be used in binary cases like this.

it's shorten

mamal72: Yeah, but it doesn't make much more sense IMO. Having more readable code is always better than having some sort of shorter code.

HelveticaScenario: > Can we have multiple types here? Being a client-side javascript, use 1/0 as a boolean is preferable

This makes no sense. I've never seen 1/0 used as a boolean anywhere else, in any language, client-side or server-side, except for C which doesnt actually have the concept of booleans so it's neccesary. It complicates an aspect of the code that has no reason to be complicated, because you have to expect truthy and falsy values, instead of just `true` and `false`. Even if this PR never gets merged you should really stop using 0 and 1 as booleans.

src/editor/config/config.ts
SamBroner: This seems like a partial solution until plugins gets typed.

SamBroner:
This is an interesting effort here. Because the framework is so configurable, I definitely would find typescript to be helpful. 

I'm not sure this is the right start though. I think it might be easier to pull back the tsconfig to a very minimal set of requirements, turn every file into .ts file and then restrict the tsconfig over time.

artf:
> I think it might be easier to pull back the tsconfig to a very minimal set of requirements

Agree

> turn every file into .ts file and then restrict the tsconfig over time.

Not sure if this is the best approach. IMHO, it will be more confusing. At least, if we see a `.js` file, we know that this file is the one to convert to `.ts` (from which I'd already expect to see TS)

jlong-crestron:
What is the status of this? for integration into larger projects with many plugins and custom modules. This type assertion is very helpful.

devtechk:
Any update for this migration to TypeScript? Thank you

artf:
I'm closing this as it's quite old and inactive as PR, but I still believe the TS migration would be a big plus for the project https://github.com/artf/grapesjs/issues/1759#issuecomment-1030758752

bostondevin:
Can you leverage the work they did on the grapesjs-react project? They made a grapesjs.d.ts file that should be something like you want: https://github.com/thanhtunguet/grapesjs-react/blob/master/libs/grapesjs-react/src/grapesjs.d.ts

PR #2227, opened 2019-08-28
Asset Manager Upload / Store file Limit

tliscomb:
asset manager added config & logic to set max file(s) size; preventing the user from exceeding their limit.

### DIFF
--- a/src/asset_manager/view/FileUploader.js
+++ b/src/asset_manager/view/FileUploader.js
@@ -108,6 +108,11 @@ export default Backbone.View.extend(
     uploadFile(e, clb) {
       const files = e.dataTransfer ? e.dataTransfer.files : e.target.files;
       const { config } = this;
+      const { beforeUpload } = config;
+
+      const beforeUploadResponse = beforeUpload && beforeUpload(files);
+      if (beforeUploadResponse === false) return;
+
       const body = new FormData();
       const { params, customFetch } = config;
 
--- a/src/asset_manager/view/FileUploader.js
+++ b/src/asset_manager/view/FileUploader.js
@@ -115,12 +115,20 @@ export default Backbone.View.extend(
         body.append(param, params[param]);
       }
 
+      var fileSize = 0;
+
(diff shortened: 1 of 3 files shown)


src/asset_manager/view/FileUploader.js
```diff
@@ -115,12 +115,20 @@ export default Backbone.View.extend(
...
+          fileSize += files[i].size;
         }
       } else if (files.length) {
         body.append(config.uploadName, files[0]);
+        fileSize += files[0].size;
+      }
+      if (config.uploadMaxFileSize > 0 && config.uploadMaxFileSize <= fileSize / 1024) {
+        alert(config.uploadMaxFileSizeAlertText || 'The file(s) you have selected to add exceed your space limit');
+        return;
       }
```
artf: IMHO, instead of those 2 options, I'd prefer something like `config.beforeUpload({ fileSize, files, ...maybeOtherOptions })` callback where the user can define its own logic and raise its own UI (we shouldn't use `alert` in the framework, if you feel the need to do it, it means you have to create a customizable callback for the user). Eg.
```js
const { beforeUpload } = config;
const beforeUploadRes = beforeUpload && 
	beforeUpload({ fileSize, files, ...maybeOtherOptions });
if (beforeUploadRes === false) return;
```

src/asset_manager/view/FileUploader.js
```diff
@@ -115,12 +115,20 @@ export default Backbone.View.extend(
         body.append(param, params[param]);
       }
 
+      var fileSize = 0;
```
artf: Use `let`

tliscomb:
@artf, I made the changes you requested; which makes this a lot cleaner and better for me too. Thank you. 

Do I need to do a new pull request or is this sufficient?

artf:
> @artf, I made the changes you requested; which makes this a lot cleaner and better for me too. Thank you.
> 
> Do I need to do a new pull request or is this sufficient?

No, it's perfect now, thank you very much

PR #2240, opened 2019-09-03
Font Awesome 5 update

senchden:
Hello. I have updated this library to use font awesome 5 for icons. Please review it and tell me whether everything is OK if you are interested. I have tested it and there were no obvious visible bugs.


artf:
Thanks for the effort @senchden but to be honest I have no interest to update to the v5 (I apologize for your time but generally you should ask if that kind of PR is welcome), as already explained in this PR https://github.com/artf/grapesjs/pull/1932 grapesjs shouldn't contain an icon library as a dependency.

I would just create a container for collecting SVG icons (+ mini default set to replace the old one) and probably some simple API to allow plugins to pull data there

senchden:
It's fine. Could I ask whether we can get this API in the foreseeable future? Is someone working on it?

artf:
I apologize, but I have no ETA for this feature and no one is actually working on it :/

kaoz70:
Although it would be nice to get rid of icon dependencies, I think this PR would benefit people already using GrapesJS in the meantime, since most developers would be using Font Awesome already and the 4.7 version is already outdated. I have this issue currently in a couple of projects where I am using FA 5 and the CSS styles dont match.

PR #2295, opened 2019-09-26
Guard against parentNode being undefined when calling toggleEvents

### LINKED ISSUE #2294: [Bug] Text components are removed with an error if they are dragged while you are editing their contents
This can replicated in the demo: https://grapesjs.com/demo.html

1. Drag in a text node
2. Double click to activate the RTE
3. Drag the text component into another position
4. Select another component

The component is removed and the following error is logged to the console:

```
Uncaught TypeError: Cannot read property 'tagName' of null
    at r.toggleEvents (ComponentTextView.js:165)
    at r.disableEditing (ComponentTextView.js:71)
```

tom-sherman:
Fixes #2294

### DIFF
--- a/src/dom_components/view/ComponentTextView.js
+++ b/src/dom_components/view/ComponentTextView.js
@@ -161,8 +161,9 @@ export default ComponentView.extend({
 
       while (el) {
         el.draggable = enable ? !1 : !0;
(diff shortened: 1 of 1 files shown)


artf:
Thanks Tom

PR #2344, opened 2019-10-18
WIP: Allow promises inside load and store

### LINKED ISSUE #2315: Allow returning promises from StorageManager `.load` and `.store` methods
This would offer a more modern and clean API than the current callback based approach.

**Proposed API**:

```javascript
editor.StorageManager.add('simple-storage', {
  load: async keys => {
    const result = await foo(keys)
    return JSON.parse(result)
  },
  async store(data) {
	const result = await bar(data)
    return result
  }
});
```

---

There are a few ways this could be implemented:

## 1. Detect if the `load`/`save` method is thenable, if so await 

This would complicate the API and code a little as users would need to make a choice between using either the callback or Promise based approach. The upside to this though is that it would not be a breaking change.

A side effect of this API change would allow mixing and matching of approaches:

```javascript
// An async load method and callback store method
editor.StorageManager.add('simple-storage', {
  load: async keys => {
    const result = await foo(keys)
    return JSON.parse(result)
  },
  store(data, clb, clbErr) {
	bar(data, err => {
      if (err) return clbErr(err)
      clb()
    })
  }
});
```

## 2. Switch to a Promise only StorageManager API

The big benefit of this is we would be encouraging a more best practice approach and the API would be far less complicated. Not having to support two slightly different APIs is also good for maintenance. Unfortunately it would be a breaking change and might be best paired with a Promise rollout across the whole library. There are other APIs that could do with supporting promises too, commands for example.

## 3. Add new methods `asyncLoad` and `asyncStore`

The user would need to choose to either use the callback-based `load`/`store` methods, or their promise-based async counterparts. To me, this seems like the worst of options 1 & 2.

I'd be happy to open a PR for this, but would be good to get thoughts on which approach is best before continuing :smile:

tom-sherman:
Closes #2315 

## `callbackOrPromise({ fn, args, success, error }) => Promise`

This function is a bit complicated but is needed to support both callbacks and promises at the same time. It always returns a promise, but the resolved data is different depending on if we use callbacks or Promises in `fn`.

If you pass callback(s) to this function, it will treat `fn` as a callback based function. This means we can choose not to pass `callbackOrPromise` the required callback options and the resulting promise will correctly resolve/reject according to the result of `fn`.

Todo:

- [x] Fix tests
- [ ] Add documentation


src/storage_manager/index.js
```diff
@@ -35,6 +35,7 @@
 import defaults from './config/config';
 import LocalStorage from './model/LocalStorage';
 import RemoteStorage from './model/RemoteStorage';
+import callbackOrPromise from '../utils/callback-or-promise';
```
artf: We use this for the module resolution in webpack
```js
resolve: {
      modules: ['src', 'node_modules'],
	  ...
```
so you can do this ` from 'utils/callback-or-promise';
`

test/main.test.js
```diff
@@ -25,4 +25,6 @@ describe('Main', () => {
   require(`${path}style_manager`);
   require(`${path}trait_manager`);
   require(`${path}grapesjs`);
+  require(`${path}grapesjs`);
```
artf: Added twice `${path}grapesjs`

test/specs/storage_manager/index.js
```diff
@@ -101,6 +103,51 @@ describe('Storage Manager', () => {
...
+        };
+        const data2 = {};
+        const id = obj.getConfig().id;
+        data2[id + 'item'] = 'testData';
+        data2[id + 'item2'] = 'testData2';
+
+        await obj.store(data);
+        const res = await obj.load(['item', 'item2']);
+        expect(storeValue).toEqual(data2);
+        expect(res).toBeUndefined();
```
artf: is it supposed to be `expect(res).toEqual(data)` or I'm missing something?! 🤔

test/specs/utils/callback-or-promise.js
```diff
@@ -0,0 +1,169 @@
+import callbackOrPromise from '../../../src/utils/callback-or-promise';
```
artf: load via `...from 'utils/callback-or-promise'`

artf:
So far I really like this PR, especially for solid tests and clean code 👍

test/specs/storage_manager/index.js
tom-sherman: Whoops, yes it should! :laughing:

tom-sherman:
@artf I'm closing this because it's cluttering my todo list 😆

PR #2345, opened 2019-10-19
Use jest runner

### LINKED ISSUE #2341: Should use jest runner to run tests
Right now we are using a single `main.test.js` file to run all of the tests. This stops Jest from running the tests in parallel, so we don't get nice progress bars and it takes ages! We also cannot run only tests relating to changed files.

We should switch to using the recommended approach of allowing jest to pickup all test files inside of the specs directory.

tom-sherman:
Closes #2341 

A massive diff but the changes are pretty simple!

- Remove all of the `module.exports = { run() {} }` in favour of using Jests runner to pick up files. This is configured in the package.json

- Skip the sorter test for now, I don't think it was running before because it doesn't use Jests syntax

- Because of the first change, I've removed a few files which only imported and ran tests in other files


artf:
> Skip the sorter test for now, I don't think it was running before because it doesn't use Jests syntax

Correct, I was planning to refactor the Sorter but then I've left it as it is

BTW I'm totally in for the recommended approach, but I tried to run it on my machine and it seems to be slower.
With the previous approach, time ranges between 40s and 60s, with the new one 50s and 110s. Is it the same on your side?

tom-sherman:
@artf Here are results running `npm t` 10 times on my machine:

  | Current | This PR
-- | -- | --
Min | 19.276 | 27.318
Max | 23.675 | 30.305
Mean | 20.494 | 28.774

Looks like using this approach is 40% slower when running the full test suite. However this PR allows you to use Jest's `o` watch mode to only run tests related to files changed since the last commit - this would speed up regular development massively. It also allows for debugging tests, something I couldn't get to work with the current test suite.

artf:
> Looks like using this approach is 40% slower when running the full test suite. However this PR allows you to use Jest's o watch mode to only run tests related to files changed since the last commit - this would speed up regular development massively. It also allows for debugging tests, something I couldn't get to work with the current test suite.

Ok, I'd say that is reasonable for what we gain. Thank you very much Tom

Issue #2372, opened 2019-11-04
Dragging from toolbar can be off (dragMode)


emilsedgh:
[Here is a working demo of the issue](https://jsfiddle.net/srvmLkt6/)

Steps to reproduce:
1. Select a component
2. Try dragging it from the toolbar

You can see that immediately the component jumps to the left.
Although it works fine if you try dragging the component by itself (and not using toolbar)

This only happens if Grapesjs is loaded from a with margin and doesn't start from `x:0`

The reason it happens is this:

In both cases of dragging from toolbar or component itself, the `tlb-move` command is called.
`tlb-move` accepts an `event` argument and the positioning of the element is calculated using `event.clientX`.

When you drag the component, sine the component is inside the iframe, it's clientX is relevant to the document and all goes fine.

But the toolbar button doesn't live on the canvas/iframe. Therefore, an event originating from `ToolbarButtonView` is relevant to the page, not to the canvas.

Here is a visual description:

![Screenshot_20191104_023411](https://user-images.githubusercontent.com/120439/68114740-a1449580-feab-11e9-8571-6f3991cbd8b8.png)

The ideal solution would be to calculate clientX and clientY of ToolbarIconView relative to the canvas before passing them to `tlb-move` command.

If you would accept such patch, I can get started on one.

pouyamiralayi:
@emilsedgh there is an open discussion about this [here](https://github.com/artf/grapesjs/issues/2207)

emilsedgh:
Oh, I tried to find it before reporting but I couldn't. Thanks. I'll close this one then.

alemenciones:
this's fixed in the last release right?

I need a solution when i drag outside of iframe in mobile devices

PR #2375, opened 2019-11-04
Calibrate the coordinates based on frame position.

### LINKED ISSUE #2207: [BUG]: Can't drag component if there is some space on top/left with Mode Absolute
Given the following code example...

``` html
<div style="height: 100px;">
  Some Space on top or left
</div>
<div id="gjs">
  <p style="padding: 25px">
    Hesdllo World!!!</p>
</div>

```

``` js
editor.getModel().set('dmode', 'absolute');
```

Results in no ability to drag and drop elements to correct position using drag icon on the panel.

image

See Fiddle for demonstration...

https://jsfiddle.net/niisar/zc51gtnv/

emilsedgh:
Calibrate the coordinates based on frame position.

Since the toolbar lives outside the canvas frame, the event's generated on it
have clientX and clientY relative to the page

This patch makes sure we consider frame offset from the page. 

Fixes #2207,#2372


emilsedgh:
@artf Any chance we could review and merge this?

emilsedgh:
I updated the branch to resolve conflicts.

emilsedgh:
Please note that the issue has been reported at least 3 times:
#2207, #2372 and #2473

src/dom_components/view/ToolbarButtonView.js
```diff
@@ -22,9 +22,32 @@ export default Backbone.View.extend({
   handleClick(event) {
     event.preventDefault();
     event.stopPropagation();
-    const { em } = this;
```
artf: I think you forgot this 😬

emilsedgh: Oh right of course.
Can you explain what's the different between `this.em` and `this.editor`?

artf: `editor` is more or less a public API layer of `em` but they both act as a Service Manager

artf:
Thanks @emilsedgh ❤️

src/dom_components/view/ToolbarButtonView.js
emilsedgh: In that case I don't like the fact that in this patch I'm using both `em` and `editor`. I think we should use either one not both.
You agree with that? Should I update my branch or something?

ramusesan:
when can expect to be pushed this changes to **npm repo**
@artf

artf:
@ramusesan I'll try to find time to release it tomorrow

ramusesan:
@artf  ok thanks, I will be waiting

alemenciones:
What happens when i change device and the component is outside the canvas? this's not working

image

artf:
> What happens when i change device and the component stay outside of canvas? this's not working maybe
> 
> image

The outline boxes keep track of the components even outside of their frames

alemenciones:
I'm talking about the calibration of coordinates x/y in the `tlb-move` event, when the componente is outside of frame, this is not working good

PR #2446, opened 2019-12-06
escape html characters when converting text node to HTML

mcammisa:
When converting templates to HTML using the `component.toHTML()` function special characters are not escaped. This produces badly formed HTML.

Steps to reproduce:
 1. Go to https://grapesjs.com/demo.html
 1. Enter `<` into a text field
 1. In the dev console run the command `window.grapesjs.editors[0].getHtml()`

Expected Output:
Properly formed HTML with `<` character escaped

Actual Output:
`<` character left untouched and broken HTML

Please let me know if I missed anything here.

Keep up the great work!

### DIFF
--- a/src/dom_components/model/ComponentTextNode.js
+++ b/src/dom_components/model/ComponentTextNode.js
@@ -10,7 +10,12 @@ export default Component.extend(
     },
 
     toHTML() {
-      return this.get('content');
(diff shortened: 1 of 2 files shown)


artf:
Great, thank you very much Matias ❤️

PR #2474, opened 2019-12-26
Improve SelectorManager and introduce component-first selection

artf:
Hi guys, this PR brings some general improvements over the SelectorManager module (responsible to show the selected components and how they'll be styled), like adding [more options](https://github.com/artf/grapesjs/pull/2474/files#diff-3a37b9ba50d737000944efa4d2e5e69e) for customization and a few layouts update. The biggest change is the introduction of `componentFirst` option that I'd like to describe more in detail.

## Default selection
The default selection behavior acts like this:
* If you select a component without classes, the selector takes the component itself as a target and all the styles are applied only to that specific component
sm-comp-target
* If the selected component has some classes, the target will be the corresponding CSS rule (eg. if the component has `class="test1 test2"`, styles will be applied to `.test1.test2 {...}` rule)
sm-rule-target
* If you need to target component-specific styles you can disable all the selectors from the list by clicking on their checkbox icon
sm-no-selectors
The biggest [critique](https://github.com/artf/grapesjs/issues/863) of this behavior is being a bit confusional from the UX point of view. When the user selects the component it expects to change to be applied only to that specific selection. This is why `componentFirst` option was added.

## Component-first selection
When you use the `componentFirst` option
```js
grapesjs.init({
	...
	selectorManager: {
        componentFirst: true,
    },
})
```
this is what happens:
* A selected component without classes acts like before, what is changed is the case of components with classes, which will target always the component itself
sm-cmp-target
Now when you change the style those will be always applied to that specific component. 
* If the selected component contains classes you will be able to transfer styles from the component to those specific selectors via a new "Sync" button
sm-sync-btn
By clicking on it, styles on the component will be cleared and added to the selectors
sm-post-sync
* If you need you can change the behavior via API in this way
```js
editor.SelectorManager.getConfig().componentFirst = true;
```
* With this approach, it's also easier for us to apply style changes in multiple selections (in the default mode, the style manager is hidden with multiple selected components)
sm-multisel
From the image above you can see selected multiple targets and also the common selectors between those components (in this case `card__content`). The sync button will apply changes to common selectors and will remove them from targets.

Worth noting, this change doesn't break the current behavior and, by default, the `componentFirst` option is disabled (at least for now). If you have any question/suggestion about this topic please reply to this PR


emilsedgh:
I really appreciate you having backward compatibility in mind and preserving the old behavior.

If I'm understanding correctly, this allows you to apply changes to a bunch if components at the same time. I can see that being useful in some cases.

Im curios as how will it behave though. 
Let's say I have 3 components with the same class. If I click on one of them, will all 3 have the blue border around them as if I've selected all 3?

If not, it may be a bit too confusing to change the style of other elements while they are not selected.

artf:
> Let's say I have 3 components with the same class. If I click on one of them, will all 3 have the blue border around them as if I've selected all 3?
If not, it may be a bit too confusing to change the style of other elements while they are not selected.

This is actually the default behavior because the editor takes as a target the CSS rule and not the Component

MatthewMariner:
This is awesome. When can this be pushed to production?

artf:
@MatthewMariner already available in https://github.com/artf/grapesjs/releases/tag/v0.15.10

jgui1129:
Hi @artf this is a very useful feature. Thank you for making the editor becoming better overtime. I am studying now the API and building stuff on top of it. Would really like to contribute soon. Cheers!

PR #2479, opened 2019-12-30
Update Components-js.md

ackvf:
This change makes more clear what is going on and why the code will not work as expected.

Even though it was already written there, I was still unsure and I have almost posted a _[Question]_. But upon re-reading what I wrote ten more times, I finally understood what you meant by `but be aware of a string conversion,`.

### DIFF
--- a/docs/modules/Components-js.md
+++ b/docs/modules/Components-js.md
@@ -68,7 +68,7 @@ As you see the editor attaches a unique ID to all components with scripts and re
 
 Keep in mind that all component scripts are executed only inside the iframe of the canvas (isolated, just like your final template), and therefore are NOT part of the current `document`. All your external libraries (eg. jQuery) are not there, but you'll see later how to manage scripted components with dependencies.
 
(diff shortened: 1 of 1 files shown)


ackvf:
I still have a question though. Can't you instead put a shim function in the window and call that from the component?

Something like this

_original code:_
```js
var myVar = 'John';

editor.BlockManager.add('test-block', {
...
  script: function () {
    alert('Hi ' + myVar);
    console.log('the element', this);
  },
...
});
```
_follows my suggestion:_

_function (as a function not string) detected, let's use a shim:_

```
globalThis['test-block'] = identifierOfTheTestBlockComponent.script
```


```html
<div id="c764"></div>
<script>
  var items = document.querySelectorAll('#c764');
  for (var i = 0, len = items.length; i < len; i++) {
    (window['test-block'].bind(items[i]))();
  }
</script>
```

artf:
Thanks @ackvf Indeed I think it's a bit clearer right now

About the function suggestion, I'll reply on the dedicated issue

PR #2485, opened 2020-01-01
Read avoidInlineStyle from config not from attributes

emilsedgh:
If you change config after the initial configuration (like grapesjs-mjml does)
it wont be reflected on editor.attributes.avoidInlineStyle.
This means ComponentView in some cases will not respect the configuration.


artf:
Good catch, thanks @emilsedgh

PR #2524, opened 2020-01-23
Canvas refactoring for Multi Frame 

artf:
Hi guys,
I've started this branch to introduce a new way to work with frames inside the canvas. This would lead to the possibility to handle multiple frames at the same time, like creating other pages and frames with different device sizes.

multiframe

Unfortunately, there is still a lot of work to do before making it usable, this process required an important refactoring of the canvas and I've tried my best to avoid any breaking change, so before going any further I'd prefer merging what is done. 
IMHO, this refactoring brings a bit more order and sense to how frames are handled and loaded so I'd prefer working/fixing stuff with this new logic.

I don't think there is a need to talk about APIs yet, there is still a lot of work to complete before releasing something stable so they would probably change easily. 
Few things I can anticipate is definitely new methods on top of the Canvas API (eg. `getFrames()`, `addFrame({ ..props })`) and, most importantly, Components can have now multiple views attached  (think about 2 frames of the same page) so, for example, to work with its DOM reference, instead of `console.log(component.getView().el)` you would do `component.getViews().forEach(view => console.log(view.el))`.

Some points to solve before making this feature possible:
* We have to solve some [memory leak issues](https://github.com/artf/grapesjs/issues/1470), where most of them are caused by uncleared DOM references and more frames will bring more DOMs to keep in memory
* Probably a new Storage strategy to handle more frames. To keep in mind that some frames are different pages (new Components and CSS collections) and others are "mirrors" (same Components/CSS but different frame sizes)

If you have any advice or concern about this path please let me know


riteshdalal:
Hi Artur,

How is it going with this initiative? Maybe this cleanup could also allow for realtime collaboration between users editing same page. 

-Ritesh

artf:
Still far from being complete, unfortunately, but yeah, the realtime collaboration it's definitely something possible and I hope to bring it in soon

riteshdalal:
Thanks for getting back to me Artur. Looking forward to it. Also, I had a
quick question for you. Would you be open to licensing some of the
functionality you have built for Grapedrop. For example, the animation
traits.

Best,
Ritesh



Best,
Ritesh Dalal

artf:
@riteshdalal yeah probably, right now is just too much coupled with the application codebase

PR #2578, opened 2020-02-17
Added German translation

chickahoona:
In response to my offer to help with translations here https://github.com/artf/grapesjs/issues/26#issuecomment-514537828


artf:
Thanks @chickahoona ❤️

PR #2589, opened 2020-02-19
Fix bug toggling preview

Skedog:
Currently the component borders are turned on when you turn the preview off, regardless of what state the borders were in before. This change checks the button's active state to ensure the button was active before turning them on.


artf:
Good catch, thanks @Skedog

PR #2602, opened 2020-02-26
making the external assets form optional

jcamejo:
Hi @artf i hope you are having a nice day.

What do you think of this small improvement? I just got a use case where i needed to have this option and i think is a good idea.

Thanks!


artf:
Yeah cool, seems like a good idea 👍, but can you check the failing tests, please?

jcamejo:
Sorry for leaving that behind, the tests are added. There are some redeclarations of the test object to make clear that the view is being initialized with the correct config

src/asset_manager/view/AssetsView.js
```diff
@@ -56,7 +56,7 @@ export default Backbone.View.extend({
     const url = input.value.trim();
     const handleAdd = this.config.handleAdd;
 
-    if (!url) {
+    if (!url || !input) {
```
artf: mmm this `!input` check doesn't make sense here, it that case it would crash at `input.value.trim()`

jcamejo:
Absolutely right, it's changed.

jcamejo:
I'm not sure if this failure has something to do with the change, it does not look like it. Should i check some other module that i'm not aware of?

Thanks again for your assistance.

artf:
> I'm not sure if this failure has something to do with the change, it does not look like it

Nope, my fault 😬

Thanks for the contribution Juan

PR #2633, opened 2020-03-09
fix: replace static bg colors with scss var

YUzhva:
Hardcoded colors do not give the possibility to customize `StyleManager` styles:
Screenshot 2020-03-09 at 11 32 18

P.S: @artf
Please, let me know if it's required to change colors around all of the `.scss` files.
I change only those two, as I'm sure, that it would not affect the application so much.

Thank you, for merging this PR 🍻

### DIFF
--- a/src/styles/scss/_gjs_inputs.scss
+++ b/src/styles/scss/_gjs_inputs.scss
@@ -91,7 +91,7 @@
 }
 
 .#{$app-prefix}field {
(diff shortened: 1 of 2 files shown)


artf:
If you see other static colors which might benefit from variables you can submit other PRs, this one seems legit 👍 thank you for the contribution

PR #2685, opened 2020-03-25
Update Components.md

0000marcell:
In the beginning of the tutorial we encourage the reader to test these commands in the demo terminal, but these commands will not work properly without these changes, i don't know if it was intended

### DIFF
--- a/docs/modules/Components.md
+++ b/docs/modules/Components.md
@@ -44,12 +44,12 @@ editor.getWrapper().append(`<div>...`);
 ::: tip
 If you need to append a component in at a specific position, you can use `at` option. So, to add a component on top of all others (in the same collection) you would use
 ```js
-component.append('<div>...', { at: 0 })
+editor.getWrapper().append('<div>...', { at: 0 })
 ```
 or in the middle
 ```js
-const { length } = component.components();
-component.append('<div>...', { at: parseInt(length / 2, 10) })
+const { length } = editor.getWrapper().components();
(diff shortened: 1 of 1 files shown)


artf:
Hi @0000marcell thanks for pointing this out. Actually the idea was to refer a generic `component` from a canvas, not only the wrapper. So, in this case, the component might be:
```js
// Select the wrapper component
const component = editor.getWrapper();

// Get the currently selected component in the canvas
const component = editor.getSelected();

// Get the image component as a result of append
const component = editor.getWrapper().append('<img .../>')[0];
```
Do you think there is a way to make it more clear somehow/somewhere?

0000marcell:
Hi @artf , i tought about that, but for someone following the code, this variable "component" just appears and we don't know what it is, i've notice that you do this throughout the documentation and it's confusing for someone just starting, i know what "editor" is because i've created this variable, i don't know what "component" is what "command" is etc..., maybe adding the variables and the comments like you mentioned would be the best solution, or maybe just a comment

artf:
@0000marcell yeah, I totally agree with you, indeed with your changes (`editor.getWrapper()`), it doesn't actually help with that issue, it's still unclear where the `component` variable comes from. So, what I'm asking you is to put what I've suggested in another section, for example, not in the tip but right in the code example above
Schermata 2020-04-02 alle 00 32 35

Can you update it so it makes sense also to you?

PR #2689, opened 2020-03-27
Fix borders not always toggling back when exiting preview. Fixes #2637

### LINKED ISSUE #2637: [Bug]: Preview mode & automatic components borders toggling
Hello again :)

Our team recently started customising the editor & removing the webpage preset, and we might have found some bugs related to the preview mode:

1. When exiting preview mode, an error can be raised if the default `options` panel has been removed, or if the `sw-visibility` button has been removed from it (from [here](https://github.com/artf/grapesjs/blob/dev/src/commands/view/Preview.js#L52), if `editor.Panels.getButton('options', 'sw-visibility')` yields `null`, cf #2589)

A proposed solution would be to loop through the panels collection (retrieved via `editor.Panels.getPanels`) to find the `sw-visibility` button, as well as adding null-checks to avoid errors.

---

2. When the `preview` & the `sw-visibility` buttons are in the same panel, the behaviour introduced by #2589 breaks, this seems to come from the [`ButtonView`'s `active` change listener](https://github.com/artf/grapesjs/blob/dev/src/panels/view/ButtonView.js#L101), which deactivates all active buttons in the panel. So if the preview is activated, it deactivates the `sw-visibility` button before running the command, which later breaks its [active check](https://github.com/artf/grapesjs/blob/dev/src/commands/view/Preview.js#L52).

I'm afraid I cannot propose a solution regarding this point 2. due to my lack of knowledge on possible side effects modifying the buttons active change listener. But I could work your suggestions in a PR if needed.

I'd be happy to open a PR, or since this is pretty closely related to #2636, fix both issues in one.

Cheers :)

mcottret:
As described by [this issue](https://github.com/artf/grapesjs/issues/2637), the component borders may not be toggled back when exiting preview mode in some cases.

This fixes this issue by adding a `shouldRunSwVisibility` flag to the Preview command, set via the Commands API on its run (before stopping the `sw-visibility` command if it's active), & used on stop to reactivate the `sw-visibility` command if needed.

Which also implies:

- Prevent toggling of component borders while preview is active
- Rely on the Commands API to detect if the `sw-visibility` command is active rather than a panel button
- Prevent resetting the `shouldRunVisibility` flag on multiple subsequent (forced) runs / stops of the preview command

### DIFF
--- a/src/commands/view/Preview.js
+++ b/src/commands/view/Preview.js
@@ -18,7 +18,14 @@ export default {
   run(editor, sender) {
     this.sender = sender;
 
-    editor.stopCommand('sw-visibility');
+    if (!this.shouldRunSwVisibility) {
+      this.shouldRunSwVisibility = editor.Commands.isActive('sw-visibility');
+    }
+
+    if (this.shouldRunSwVisibility) {
+      editor.stopCommand('sw-visibility');
+    }
+
     editor.getModel().stopDefault();
 
     const panels = this.getPanels(editor);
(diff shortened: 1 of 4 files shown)


artf:
Awesome @mcottret thank you very much ❤️

PR #2694, opened 2020-04-01
Update color picker value when user changes value from input. Fix #2683

### LINKED ISSUE #2683: [BUG] When cancel Colorpicker, it set wrong value for component
1. **Are you using the latest release (older versions are NOT supported)?** 
    Yes.
1. **Are you facing the bug with your local copy of GrapesJS or with the current demo?**
    Both of them.
1. **Steps to reproduce:**

   * Drag a component then change its background color from input
   * Open color picker immediately
   * Then click to outside color picker (anywhere except canvas)

1. **What is the expected behavior?**
   Background color should keep remain.
1. **What happens instead?**
   Background color change to previous color or #000000.
1. **Screencast**
   
    grapesjs-backgroundcolor

longdoan7421:
As I described in issue #2683, if user changes color value from input then opens color picker and click to outside, the component will be switched back to previous color or default color (#000000).

This commit attempts to synchronize the value from input and the value from color picker. I added another condition that allow color picker update its value when value from input changed.

### DIFF
--- a/src/domain_abstract/ui/InputColor.js
+++ b/src/domain_abstract/ui/InputColor.js
@@ -44,7 +44,7 @@ export default Input.extend({
     colorEl.get(0).style.backgroundColor = valueClr;
 
     // This prevents from adding multiple thumbs in spectrum
-    if (opts.fromTarget) {
+    if (opts.fromTarget || (opts.fromInput && !opts.avoidStore)) {
       colorEl.spectrum('set', valueClr);
       this.noneColor = value == 'none';
     }
--- a/src/style_manager/view/PropertyView.js
+++ b/src/style_manager/view/PropertyView.js
@@ -76,25 +76,17 @@ export default Backbone.View.extend({
     // changes based on other properties are propagated
     const requires = model.get('requires');
     requires &&
-      Object.keys(requires).forEach(property => {
+      Object.keys(requires).forEach((property) => {
         em && em.on(`component:styleUpdate:${property}`, this.targetUpdated);
       });
 
-    this.listenTo(
-      this.propTarget,
-      'update styleManager:update',
-      this.targetUpdated
-    );
+    this.listenTo(this.propTarget, 'update styleManager:update', this.targetUpdated);
     this.listenTo(model, 'destroy remove', this.remove);
     this.listenTo(model, 'change:value', this.modelValueChanged);
     this.listenTo(model, 'targetUpdated', this.targetUpdated);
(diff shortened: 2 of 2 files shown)


artf:
Hi @longdoan7421 really cool, I was aware of the bug but never find time to debug it, so thanks for the PR.
Actually, I've tried to use the color picker from the text shadow property and only this seems to work
```js
if (opts.fromTarget || !opts.avoidStore) {
```
can you confirm that it works also in your case?

artf:
sorry, my use case was a bit different:
* I select the component
* add a new layer from the text-shadow and change its shadow color
* select another component and then select again the previous one
* now, if I try to open the colorpicker of the text-shadow the color is black

longdoan7421:
I understand your case now, I will investigate that bug more.

longdoan7421:
After researching, I think maybe this is the cause:
https://github.com/artf/grapesjs/blob/ce81833350dbd0beeeff4766760376bc9d928ab1/src/style_manager/view/PropertyView.js#L634

I don't see the option `targetUpdate` was used in anywhere. So I try to change it to `fromTarget` and it seems to work.

artf:
@longdoan7421 yep, seems like that was the point 😄

artf:
if you update the PR, I'll merge it :)

longdoan7421:
@artf Updated PR :relaxed:

src/domain_abstract/ui/InputColor.js
```diff
@@ -44,7 +44,7 @@ export default Input.extend({
     colorEl.get(0).style.backgroundColor = valueClr;
 
     // This prevents from adding multiple thumbs in spectrum
-    if (opts.fromTarget) {
+    if (opts.fromTarget || (opts.fromInput && !opts.avoidStore)) {
```
artf, line 47: We don't need this anymore :)

longdoan7421: Actually, it fixes another bug which I described in issue #2683. 
That bug is different from your bug. It happens when user changes color from input, not from color picker.
I also attempt to revert that line, but that bug is still happening.

artf: Ah, is it still happening on your side? I've tried with the same bg gradient plugin and seems to be fixed

artf: ah ok nvm, you're right, I didn't try the input thing 👍

PR #2695, opened 2020-04-01
closing the color picker if click event happens on canvas, saving the current value

jcamejo:
Hi @artf, I hope are you well.

This is the fix for the color picker problem described some time ago here:

https://github.com/artf/grapesjs/issues/2467

Let me know if i can attach the hide function to a different event, at the moment i think is the best approach.

Thanks!


artf:
Great, thank you very much @jcamejo 
By the way, can you reach me out on https://discord.gg/QAbgGXq, please?

PR #2756, opened 2020-05-07
Checking if input is active in the main document as well as the frame.

### LINKED ISSUE #2422: [BUG] Unable to add 'space' character if zoom is applied on canvas
Cannot edit text components properly if a zoom value is applied on the canvas.

GrapesJS version:0.15.8

The issue can be reproduced in the demo as well:
[https://jsfiddle.net/1n9ruwa4/](https://jsfiddle.net/1n9ruwa4/)

**Expected behavior:**
Zoom is applied on canvas using:editor.Canvas.setZoom()
The text components can be still edited and 'space'(Keycode:32) characters added

**Current behavior:**
Zoom is applied on canvas using:editor.Canvas.setZoom()
Even though other characters can be added/edited, you are not able to add new 'space' characters since the space character is also used for 'repositioning' canvas if zoom is applied.
The grab cursor style is shown and you could drag the canvas by 'space + drag'

**Steps to reproduce:**
Open jsfiddle demo above(with editor.Canvas.setZoom(90))
Try to edit the hello world text
Try to add a 'space' character

The size of the zoom level does not matter, the issue happens with any value.

mattdeigh:
Closes #2422

This checks if there is an active input in the main document as well as the frame. The onKeyPress function was trapping space bar presses in the editor and in the main document when the canvas was not at 100% zoom.

### DIFF
--- a/src/canvas/index.js
+++ b/src/canvas/index.js
@@ -552,8 +552,12 @@ export default () => {
      */
     isInputFocused() {
       const doc = this.getDocument();
+      const frame = this.getFrameEl();
       const toIgnore = ['body', ...this.getConfig().notTextable];
-      const focused = doc && doc.activeElement;
+      const docActive = frame && document.activeElement === frame;
+      const focused = docActive
+        ? doc && doc.activeElement
+        : document.activeElement;
 
       return focused && !toIgnore.some(item => focused.matches(item));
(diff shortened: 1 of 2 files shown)


artf:
Awesome, thank you Matthew

Issue #2779, opened 2020-05-15
[Fix] French i18n translations fixes


bt0r:
# Description
Hello guys, 

`Padding` translations keys are wrongly translated in French language.
As Padding is an english word, there is no sense to keep it in French. 

# Correction
Fixes can be differents depending on which kind of users you want to reach. 
For a developer, Padding & Margin are really common but if its for a more "standard" user, you should translate margin and padding as internal and external margin.

#### Actual
File : https://github.com/artf/grapesjs/blob/dev/src/i18n/locale/fr.js#L97-L106
```JSON
margin: 'Marge',
      'margin-top': 'Marge supérieure',
      'margin-right': 'Marge droite',
      'margin-left': 'Marge gauche',
      'margin-bottom': 'Marge inférieure',
      padding: 'Padding',
      'padding-top': 'Padding supérieur',
      'padding-left': 'Padding gauche',
      'padding-right': 'Padding droite',
      'padding-bottom': 'Padding inférieur',
```

#### Must be
```JSON
margin: 'Marge externe', // Means External Margin
      'margin-top': 'Marge externe supérieure', // Top external margin  etc..
      'margin-right': 'Marge externe droite',
      'margin-left': 'Marge externe gauche',
      'margin-bottom': 'Marge externe inférieure',
      padding: 'Marge interne', // Internal margin
      'padding-top': 'Marge interne supérieure', // Top internal margin
      'padding-left': 'Marge interne gauche',
      'padding-right': 'Marge interne droite',
      'padding-bottom': 'Marge interne inférieure',
``` 

Hope it can help :wave:

PR #2780, opened 2020-05-15
[I18N] Fix margin/padding keys for french translations

bt0r:
# Description
As described here #2779 , margin and padding are not well translated.
This PR edit padding/margin translations to have a notion of internal/external margin logics.


artf:
Thanks @bt0r 👍

Issue #1470, opened 2018-09-28
[Question] Possible memory leaks


mgotovkin-plana:
Hi, I am working on a project with Angular 6 and I am investigating a memory leak after destroying the editor on **_previousAttributes** and possibly coming from underscore **bind** function
![image](https://user-images.githubusercontent.com/35450180/46211535-6ed97800-c33c-11e8-9097-1636017f24c0.png)
So,
1. Why not using native bind and other functions?
2. I figure out that there are couple of underscore libraries inside grapes.js file 1.8.3 and 1.9.1, I guess caused by backbone-undo package, do you have any plans on fix this?

Thanks,
Regards Max

artf:
@BrainMaxxx would be glad to have a help on this :)

forssri:
This problem still exists, even after invoking the editor.destroy() method it is not freeing up the memory. In scenarios where multiple instances of grapesjs editors are used.

varadero:
Another possible memory leak caused by `editor.load()` is described in https://github.com/artf/grapesjs/issues/2403

simplecommerce:
Here is a small example to reproduce the issue I believe.

https://codesandbox.io/s/react-router-basic-49ki1

If you click `Dashboard`, you will see the editor, if you click any other links and go back to the `Dashboard` and take a look at the memory, it keeps going up and up, even if I use destroy in the unmounting of the dashboard component.

artf:
I apologize but this is not yet solved. I've discovered that after removing the component the reference to its DOM was still there. Unfortunately, removing the reference brokes the Sorter, so I have to figure out how to refactor it properly

artf:
Just a quick update, I'm working on this issue [here](https://github.com/artf/grapesjs/tree/fix-leaks-2), so far memory leaks with components seems to be fixed, now I'm trying to solve the one with destroying part (a bit longer as I have to isolate each module, a bit of mess in each of them...)

0000marcell:
Hi @artf I'm block because of a problem that i think is related to this, if you select a element in the canvas then run editor.destroy() and resize the window, the whole application breaks because the listeners are not being removed, stopListening is being called, but aparently is not removing the window resize listener declared in 'view/SelectComponent' toggleSelectComponent

artf:
@0000marcell in the next release, I'll probably include some fixes to avoid major memory leaks inside components but this issue is still far from being closed.
For now, if you know exactly where the leak is, I'd be happy to merge a PR

[...]

PR #2788, opened 2020-05-18
editor.destroy not removing event listeners created by SelectComponent

0000marcell:
If you select any component in the canvas, run editor.destroy() then resize the window or scroll, the app will call onFrameUpdated and onFrameScroll but the editor doesn't exist anymore which makes the app crash, related to #1470


0000marcell:
@artf Sorry, didn't have the time to test it yet, gonna check it out this week :+1:

PR #2872, opened 2020-07-02
Expose options parameter to `setStyle` & `setComponents`. Closes #2856

### LINKED ISSUE #2856: [Feature Request]: Expose options parameter to `setStyle` & `setComponents`
Hello again !

Our team has recently been trying to update GrapesJS' canvas programmatically **without accidentally triggering an `update` event**. If our understanding is correct, this would mean allowing a call to `setStyle` or `setComponents` not to trigger an `update` event, regardless of the current `stepsBeforeSave` & `changesCount` values.

After some research, the `avoidStore` option seems like a good fit to achieve what we're trying to do, but the problem is that we didn't find any way to use this parameter with the current public API.

A potential solution we found would imply exposing the options parameter to the [`setStyle`](https://github.com/artf/grapesjs/blob/dev/src/editor/index.js#L306) & [`setComponents`](https://github.com/artf/grapesjs/blob/dev/src/editor/index.js#L260) methods, to be able to do the following:

```javascript
const options = {avoidStore: true};
this.editor.setStyle([/*...*/], options);
this.editor.setComponents([/*...*/], options);
```

Do you consider this a valid use case, or is there some other way to achieve this that we could not find ?

As always, I'd be happy to open a PR if so !

Cheers :)

mcottret:


artf:
🔝 Thanks

PR #2879, opened 2020-07-06
Fix retrieval of layers' visibility always returning `true`. Fixes #2863

### LINKED ISSUE #2863: [Bug]: Programmatic retrieval of layers' visibility always returns `true`
Hello again :)

Our team might have found a bug while using the Layer Manager API: every call to `component.viewLayer.isVisible()` seems to always return `true`, regardless of their actual state.

Steps to reproduce:

- Open the [demo](https://grapesjs.com/demo.html)
- Switch to the layers panel
- Click on the eye icon to hide the "Body" layer
- Open the console
- Executing `editor.getWrapper().viewLayer.isVisible()` returns `1` instead of `false`

Looking at [the code](https://github.com/artf/grapesjs/blob/dev/src/navigator/view/ItemView.js#L303), I believe `this.model.getStyle()` should be used instead of `this.model.get('style')` (which always returns `{}`) to retrieve the current `display` value.

As always, I'd be happy to open a PR to fix it !

Cheers :)

mcottret:
\+ Remove unused import

### DIFF
--- a/src/dom_components/view/ComponentView.js
+++ b/src/dom_components/view/ComponentView.js
@@ -4,7 +4,7 @@ import Components from '../model/Components';
 import ComponentsView from './ComponentsView';
 import Selectors from 'selector_manager/model/Selectors';
 import { replaceWith } from 'utils/dom';
-import { setViewEl, getElRect } from 'utils/mixins';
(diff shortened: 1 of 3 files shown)


artf:
Great, thank you @mcottret

PR #2881, opened 2020-07-08
Made Helvetica sans-serif by default

### LINKED ISSUE #2880: BUG: Helvetica font trait should be sans-serif
**Version:**
0.16.18

**Are you able to reproduce the bug from the demo?**
Yes

The "Helvetica" font should be selected when loaded.

Since [Helvetica is a sans-serif](https://en.wikipedia.org/wiki/Helvetica#:~:text=Helvetica%20or%20Neue%20Haas%20Grotesk,sans-serif) font, it should be defined as one for the typography selector.

Currently, it is not defined as a sans-serif font, only serif, so when loading in a font-family of "Helvetica, sans-serif", the Font selector in the Typography trait is not selected as it should be

**Are you able to attach screenshots, screencasts or a live demo?**
Yes

https://codesandbox.io/s/infallible-tdd-q5ys3

adriangroch:
Fixes #2880

### DIFF
--- a/src/style_manager/model/PropertyFactory.js
+++ b/src/style_manager/model/PropertyFactory.js
@@ -548,7 +548,7 @@ export default () => ({
             'Comic Sans MS, cursive' + ss,
             'Courier New, Courier, monospace',
             'Georgia, serif',
-            'Helvetica, serif',
+            'Helvetica' + ss,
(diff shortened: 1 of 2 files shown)


artf:
Good catch, thanks @adriangroch

PR #2907, opened 2020-07-21
Fully clear hover state on mouseleave

neon12345:
"onOut" does not fully clear the internal state for hover elements.
This can result in drawing a previous hover element after mouseleave.

To reproduce this in the demo:
Open the demo with a browser in a smaller window, hover some items and drag resize the browser window.
The last hover item is drawn again after the resize.


artf:
Great @neon12345 thanks

PR #2923, opened 2020-07-27
Not show property name on Style Manager when it has an icon

cartifon:
When I added an icon on the property, it was showing the icon and the label. This fix should only show the icon when it is defined.


artf:
I don't understand why your PR is so messed up (eg. shows updates of 19 files that are already there).  Can't really get what are your changes...
Seems like you've added your self to other commits, what kind of merge did you make?

cartifon:
To be honest, I didn't see that, Let me do it again.

artf:
Anyway @cartifon by reading your description, what about putting an empty label? 
In that case, if I want the label and the icon is not a problem (probably someone already uses this approach and the change might break it)

artf:
Hey @cartifon I'd close this, for now, let me know if my suggestion might work in your case

PR #2935, opened 2020-08-03
maping through nested rules

makbari:
This is integration of the nested rules. When nested at rules is parsed from grapesjs-parset-postcss, grapes js resolve them. 
an example is also provided in the postcss repo.

### DIFF
--- a/src/code_manager/model/CssGenerator.js
+++ b/src/code_manager/model/CssGenerator.js
@@ -107,11 +107,15 @@ export default Backbone.Model.extend({
     const selectorStrNoAdd = rule.selectorsToString({ skipAdd: 1 });
     const selectorsAdd = rule.get('selectorsAdd');
     const singleAtRule = rule.get('singleAtRule');
+    let isGjsRule = false;
     let found;
 
     // This will not render a rule if there is no its component
     rule.get('selectors').each(selector => {
       const name = selector.getFullName();
+      if (name.startsWith('.gjs')) {
+        isGjsRule = true;
+      }
       if (
         this.compCls.indexOf(name) >= 0 ||
         this.ids.indexOf(name) >= 0 ||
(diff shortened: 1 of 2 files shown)


src/code_manager/model/CssGenerator.js
```diff
@@ -107,11 +107,15 @@ export default Backbone.Model.extend({
     const selectorStrNoAdd = rule.selectorsToString({ skipAdd: 1 });
     const selectorsAdd = rule.get('selectorsAdd');
     const singleAtRule = rule.get('singleAtRule');
+    let isGjsRule = false;
```
artf, line 110: what is the point of this variable? Seems like you don't use it...

src/parser/model/BrowserParserCss.js
```diff
@@ -15,7 +15,7 @@ const atRules = {
 };
 const atRuleKeys = keys(atRules);
 const singleAtRules = ['5', '6', '11', '15'];
-const singleAtRulesNames = ['font-face', 'page', 'counter-style', 'viewport'];
+const singleAtRulesNames = ['font-face', 'counter-style', 'viewport'];
```
artf, line 18: Why did you remove the `page`?

src/code_manager/model/CssGenerator.js
```diff
@@ -121,8 +125,10 @@ export default Backbone.Model.extend({
       }
     });
 
-    if ((selectorStrNoAdd && found) || selectorsAdd || singleAtRule) {
-      const block = rule.getDeclaration();
+    if ((selectorStrNoAdd && found) || selectorsAdd || singleAtRule || rule.attributes.style) {
```
artf, line 128: `rule.attributes.style` is an object, it's a truthy value, so the condition is pointless

src/code_manager/model/CssGenerator.js
```diff
@@ -121,8 +125,10 @@ export default Backbone.Model.extend({
...
+      const entries = Object.entries(rule.attributes.style);
+      const styleStr = entries.length > 0 ? entries.map(([k, v]) => `${k}:${v}`).join(';') + ';' : '';
+      const block = rule.getDeclaration().length > 0 ? rule.getDeclaration() : styleStr;
```
artf, line 131: You're calling rule.getDeclaration() twice and honestly, I don't understand the point of this change

artf (changes requested):
Can you please explain the point of this PR? What are you trying to solve?

makbari:
Thanks for the review. I will make some correction and more updates. 
There problem I am trying to solve is that, page at rule, can have nested at rule. 
The example is also in https://github.com/artf/grapesjs-parser-postcss/pull/1

PR #2975, opened 2020-08-26
handle case when url is not a base64 format .

meyerco:
- [x] add quotes arround url() method parameter .


artf (approved):

artf:
Great, thanks @meyerco

PR #2997, opened 2020-09-03
WIP: Feature - disable javascript

kuhelbeher:
@artf I need help here.

In this issue - https://github.com/artf/grapesjs/issues/2972#issuecomment-682261921 you mentioned that for disabling javascript Frame should be reloaded with current components state. Could you pls give a hint how to do it :)

### DIFF
--- a/src/commands/index.js
+++ b/src/commands/index.js
@@ -62,7 +62,8 @@ export default () => {
     ['component-exit', 'ComponentExit', 'select-parent'],
     ['component-delete', 'ComponentDelete'],
     ['component-style-clear', 'ComponentStyleClear'],
-    ['component-drag', 'ComponentDrag']
+    ['component-drag', 'ComponentDrag'],
+    ['disable-js', 'DisableJs', 'disable-js']
   ];
 
   // Need it here as it would be used below
--- a/src/panels/config/config.js
+++ b/src/panels/config/config.js
@@ -1,3 +1,4 @@
+const djs = 'disable-js';
 const swv = 'sw-visibility';
 const expt = 'export-template';
 const osm = 'open-sm';
@@ -19,6 +20,14 @@ export default {
     {
       id: 'options',
       buttons: [
+        {
+          active: true,
+          id: djs,
+          className: 'fa fa-file-code-o',
+          command: djs,
+          context: djs,
+          attributes: { title: 'Disable JS' }
+        },
         {
           active: true,
           id: swv,
--- a/index.html
+++ b/index.html
@@ -1,9 +1,9 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
   <head>
-    <meta charset="utf-8">
+    <meta charset="utf-8" />
     <title>GrapesJS</title>
-    <link rel="stylesheet" href="dist/css/grapes.min.css">
+    <link rel="stylesheet" href="dist/css/grapes.min.css" />
     <script src="dist/grapes.min.js"></script>
     <style>
       body,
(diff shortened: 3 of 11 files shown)


src/commands/index.js
```diff
@@ -62,7 +62,8 @@ export default () => {
     ['component-exit', 'ComponentExit', 'select-parent'],
     ['component-delete', 'ComponentDelete'],
     ['component-style-clear', 'ComponentStyleClear'],
-    ['component-drag', 'ComponentDrag']
+    ['component-drag', 'ComponentDrag'],
+    ['disable-js', 'DisableJs', 'disable-js']
```
artf, line 66: The third option is used to support old commands, so you don't need it here :)

src/panels/config/config.js
```diff
@@ -19,6 +20,14 @@ export default {
...
       id: 'options',
       buttons: [
+        {
+          active: true,
+          id: djs,
+          className: 'fa fa-file-code-o',
+          command: djs,
+          context: djs,
+          attributes: { title: 'Disable JS' }
+        },
```
artf, line 30: You can keep this button for the development purpose but in the end, I wouldn't include it

artf:
Good start Dmitriy.
Basically you have added the `enableJs` property in the Canvas model (propably I'd call it `js` for the sake of brevity), now in its `initialize` set a listener and update the same (new) property in its frame models.
```js
/// model/Canvas.js
initialize() {
 //...
 this.on('change:js', this.onJsChange);
}
onJsChange() {
 const js = this.get('js');
 this.get('frames').forEach(frame => frame.set({ js }))
}
```
The next step is to re-render the frame on its property change
```js
// view/FrameView.js
// ...
initialize(o) {
 //...
 this.listenTo(this.model, 'change:js', this.rerender); //
}
rerender() {
 // clean current wrapper
 this.render();
}
```
I never thought about rerendering the frame so I can't tell you what side effects might come out (before triggering the render you should remove/clean the actual wrapper view so all its listeners will be removed)

src/commands/index.js
kuhelbeher: I tried to remove 3rd option, but I receive this warning `'disable-js' command not found` on running this command via panel button.

kuhelbeher:
> I never thought about rerendering the frame so I can't tell you what side effects might come out (before triggering the render you should remove/clean the actual wrapper view so all its listeners will be removed)

What do you mean by "remove/clean the actual wrapper"? You mean FrameWrapView? And how can I do it?

I tried to remove FrameWrapView and FrameView (with remove method) but I haven't be able to restore them.

artf:
> What do you mean by "remove/clean the actual wrapper"? You mean FrameWrapView? And how can I do it?

No sorry, I'm talking about the actual [Wrapper Component](https://github.com/artf/grapesjs/blob/dev/src/dom_components/model/ComponentWrapper.js) (the root component)


> I tried to remove FrameWrapView and FrameView (with remove method) but I haven't be able to restore them.

No, those shouldn't be removed as also we don't need to touch Component's models (otherwise a change/undo/etc. will be triggered). 
We need to re-render the iframe itself (FrameView), so this will clean eventually all the objects/events/instances created inside the iframe window and, re-render the Wrapper Component (so this will generate new View Components for each Model Component)

kuhelbeher:
Sorry for bothering you again. I tried to [remove](https://github.com/artf/grapesjs/blob/dev/src/dom_components/view/ComponentView.js#L89) and then [render](https://github.com/artf/grapesjs/blob/dev/src/dom_components/view/ComponentView.js#L508) again Wrapper Component but I hit the error after removing it:
#### **`FrameView.js`**
```javascript
rerender() {
  this.root.remove();
  this.root.render(); // TypeError: Cannot read property 'attributes' of undefined

  this.render();
}
```
Also I tried to [reset](https://github.com/artf/grapesjs/blob/dev/src/dom_components/view/ComponentView.js#L459) the Wrapper:
#### **`FrameView.js`**
```javascript
rerender() {
  this.root.reset();

  this.render();
}
```
But in this case nothing happened and after reset I couldn't drop new blocks to canvas anymore.

Could you please give some hint?

artf:
> I never thought about rerendering the frame so I can't tell you what side effects might come out

Sorry but this is exactly what I was talking about. There is probably a need to add a new method in ComponentView which rerenders the view in a proper way and without side effects

kuhelbeher:
Could you please point out on what edge cases and side effects should pay attention on rerendering the ComponentView?

I'm not very familiar with grapesjs source code but I will try to add it.

artf:
Unfortunately, it's hard to tell, you need to ensure that when the view is removed all its events are cleared, to avoid also memory leaks. 
Honestly, I believe it's quite a hard task to complete without a deep knowledge of how components (model and view) work, but I can't really help you more on this, I have other priorities right now.

artf:
@kuhelbeher I'm closing this one as I don't think the PR solves exactly the issue in the proper way (due also to some new changes in the latest releases) but I'm still keeping the related issue open and I hope to come up with a solution soon.

PR #3029, opened 2020-09-16
Fix non integer steps for slider property

kuhelbeher:
Fixes artf/grapesjs#2748


artf (approved):

artf:
Great, thanks Dmitriy

PR #3158, opened 2020-11-25
Add keydown event to handle up and down arrows keypress events

kuhelbeher:
Added feature for inputNumber where user can change input value using up and down arrows, similar to up and down arrows to the right of the number input.


artf (approved):

artf:
Nice 👍  thanks

PR #3182, opened 2020-12-11
fix SelectorManager not reutrn state string

congweibai:
I think could fix the issue I reported here https://github.com/artf/grapesjs/issues/3181


artf (approved):

artf:
Makes sense to me 👍  Thanks

PR #3220, opened 2021-01-06
Change units on resizing to pixels

### LINKED ISSUE #3186: BUG: Resize with percentage
Hello, there is a bug with resizing elements when their sizes are set with percentage:
simplescreenrecorder-2020-12-11_15 46 17

You can check this issue [here](https://codesandbox.io/s/grapes-resize-2ycel?file=/src/index.js)

kuhelbeher:
Here is my naive attempt to fix #3186

This is working, but I may be missing something. @artf please, could you check it?

### DIFF
--- a/src/utils/Resizer.js
+++ b/src/utils/Resizer.js
@@ -18,7 +18,7 @@ var defaultOpts = {
   step: 1,
 
   // Minimum dimension
-  minDim: 32,
+  minDim: 10,
 
   // Maximum dimension
   maxDim: '',
--- a/src/utils/Resizer.js
+++ b/src/utils/Resizer.js
@@ -458,14 +458,14 @@ class Resizer {
     const maxDim = opts.maxDim;
     const deltaX = data.delta.x;
     const deltaY = data.delta.y;
-    const startW = startDim.w;
-    const startH = startDim.h;
     const parentW = this.parentDim.w;
     const parentH = this.parentDim.h;
-    const currentPosX = this.currentPos.x;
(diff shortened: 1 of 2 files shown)


artf:
I think the proper way would be to leave the unit selected by the user, so it should make the px calculation during the drag and then convert it back to the original unit. While I'm dragging, I'd still expect to see the original unit in style manager (so the formatting should happen at any drag move)

kuhelbeher:
@artf hello, sorry for long response. I've changed formulas for calculating element dimensions based on selected unit. Could you check it pls?

Also I have a question about minDim property - it was 32 by default, but for percentages it's too big restriction. I simply reduced to 10, but maybe you have better idea?

src/utils/Resizer.js
```diff
@@ -441,8 +458,14 @@ class Resizer {
...
-    const startW = startDim.w;
-    const startH = startDim.h;
+    const parentW = this.parentDim.w;
+    const parentH = this.parentDim.h;
+    const unitWidth = this.opts.unitWidth;
+    const unitHeight = this.opts.unitHeight;
+    const startW =
+      unitWidth === '%' ? (startDim.w / 100) * parentW : startDim.w;
+    const startH =
+      unitWidth === '%' ? (startDim.h / 100) * parentH : startDim.h;
```
artf, line 468: it should be `unitHeight` here but it's ok, I'll take care of it

artf (approved):

artf:
Tried it locally, works super well, thanks @kuhelbeher ❤️

PR #3229, opened 2021-01-12
right variable validation

jcamejo:
Hi!

Checking the releases I've noticed that i wrote the wrong boolean in the second validation.


artf (approved):

artf:
Thanks Juan

Issue #3267, opened 2021-02-11
BUG: Cannot set 0px box shadow blur


tudor-ooo:
For some reason, the editor does not allow to set the blur to 0px.
0 always magically turns into 5px.

<img width="243" alt="Screenshot 2021-02-11 at 17 14 05" src="https://user-images.githubusercontent.com/668451/107664461-ba321e00-6c8c-11eb-91ce-4e8a87a2e33c.png">

**Are you able to reproduce the bug from the demo?**
Yes

You can find this bug here:
https://grapesjs.com/demo.html

artf:
Thanks, I guess there is some falsy check with the `0` which leads to the default 5px value.

akankshach29:
Would like to help fix this bug. Will raise a PR.

artf:
This one should already be fixed

PR #3297, opened 2021-03-01
fix: box shadow blur default value

akankshach29:
This is to fix the bug: Cannot set 0px box shadow blur. Reference: #3267


artf:
Hi @akankshach29 thanks for the PR, but unfortunately, this is not how I'd like to solve it. I mean, I'd like to keep the default `5px` for the blur as visually is way more readable. 
Probably, the main issue to fix here is somewhere in style properties where `0` is not accepted properly.
Let me know in case you need more details.

PR #3300, opened 2021-03-01
fix to script saving to be resilient if scripts contain comments

abulka:
If a component script has javascript comments and is saved, there is a high risk of `Uncaught SyntaxError: Unexpected end of input` in `src/canvas/view/CanvasView.js:312` and the script will not be exported properly.

## Repro

Saving a script with comments e.g.
```javascript
let el = this // hi
```
or
```javascript
//let el = this
```

Digging deeper it looks like grapesjs is trying to append a script object with innerHTML of:

```javascript
innerHTML: "
      setTimeout(function() {
        var item = document.getElementById('i6x8e');
        if (!item) return;
        (function(){//let x=1333;}.bind(item))({})
      }, 1);"
```

and the comment inadvertently blows away the `.bind(...` portion of the script which was attached by Grapesjs.

Issue first discovered and described at https://github.com/Ju99ernaut/grapesjs-script-editor/issues/4

## The fix

By inserting newlines around the user script, we ensure comments will not blow away portions of the script which are attached by Grapesjs.  Viz

```javascript
innerHTML: "
      setTimeout(function() {
        var item = document.getElementById('i6x8e');
        if (!item) return;
        (function(){
//let x=1333
;}.bind(item))({})
      }, 1);"
```

## Reservations

- Not sure about the trailing `;` being added by Grapesjs?
- Tests not run.
- This is my first contribution to Grapesjs, so not sure about the deeper implications of this change. 
- Whilst I had to add the newlines fix in `src/code_manager/model/JsGenerator.js` to `build()` too, there remains another line of code in build() that probably needs it too line 64: `(${mapType.code}.bind(el))(props[el.id]);` but I haven't struck the need for it quite yet, so have left it.

### DIFF
--- a/src/canvas/view/CanvasView.js
+++ b/src/canvas/view/CanvasView.js
@@ -299,7 +299,7 @@ export default Backbone.View.extend({
     const scriptFn = model.getScriptString();
     const scriptFnStr = model.get('script-props')
       ? scriptFn
-      : `function(){${scriptFn};}`;
+      : `function(){\n${scriptFn}\n;}`;
     const scriptProps = JSON.stringify(model.__getScriptProps());
     script.innerHTML = `
       setTimeout(function() {
--- a/src/code_manager/model/JsGenerator.js
+++ b/src/code_manager/model/JsGenerator.js
@@ -69,7 +69,7 @@ export default Backbone.Model.extend({
         code += `
           var items = document.querySelectorAll('${ids}');
           for (var i = 0, len = items.length; i < len; i++) {
-            (function(){${mapType.code}}.bind(items[i]))();
+            (function(){\n${mapType.code}\n}.bind(items[i]))();
           }`;
       }
     }


artf (approved):

artf:
Thanks Andy, I appreciate your help.
Well, I guess that should solve the issue quite simply 👏
Anyway, I'd highly recommend using the new approach for components with js, as described here https://grapesjs.com/docs/modules/Components-js.html#passing-properties-to-scripts
It's way more reliable than the old one and the function string is not altered like before (which is the main reason of this bug).

artf (approved):

abulka:
> Anyway, I'd highly recommend using the new approach for components with js, ...
> It's way more reliable than the old one and the function string is not altered like before (which is the main reason of this bug).

Ok I just tried that `script = function (props)` approach. Interestingly, it forced me to declare some `'script-props'` otherwise `updateScript(view)` crashes with a `Function statements require a function name` error due to 
```javascript
(function(){
function (props) {
```
being generated. If I really don't have any `'script-props'` then I'll have to `'script-props': [],` to avoid that error. I mention this because if developers are being encouraged to use this newer approach but don't always need props perhaps Grapesjs could be more resilient if there are no script-props declared. Just a thought.

## User Scripts?

The reason I ran into this comments in scripts issue is because I am building a simple scripting IDE in Grapesjs. I let users add user scripts to components and don't let them see or edit component scripts - which would be confusing and which users could mess up. Those user scripts need to be able to contain comments. I have been using a popup user script editor on each component and simply appending user scripts with something like
```javascript
target.set('script', original_component_script + '\n' + user_script)
```

Whilst powerful and much needed, the `script = function (props) {...` approach makes it more challenging to inject 'user scripts' because `target.getScriptString()` looks like `function (props) { COMPONENT CODE }` which is impossible to simply append user code to - unless I perhaps do some parsing and inject the code *before* that last `}`. With the older `script = function () {...` approach `target.getScriptString()` gave me a clean script body (no function or curly braces) to which I could easily append to.

Perhaps 'user scripting' should be done differently? Perhaps I could implement a component user script as a separate property and invoke it explicitly somehow by the particular component script? Or perhaps a user_script property could be officially supported by Grapesjs at a deeper level, and properly generated during export together with normal scripts. 

Updating component scripts is also an issue. When their definition changes (component maker updates the component) existing components on the canvas never get updated since they have a copy of the component script coming from persistence which takes precedence (whenever a user script has been appended, and I think `component.get('scriptUpdated')` is true). Again this suggests a deeper Grapesjs architectural approach might be needed?

Would appreciate any thoughts - am happy to create a FEAT issue for this if you think that's the best way forward. Just as happy to accept any advice on how to implement user scripts without needing to change Grapesjs!

abulka:
> Perhaps I could implement a component user script as a separate property and invoke it explicitly somehow by the particular component script?

It turns out that this approach works. FYI this is how I implemented it: 

```javascript
const script = function (props) {
    ...
    if (props['userscript'] && (!document.getElementById("wrapper"))) // exported mode
        eval(`(function(){\n${props['userscript']}\n}).bind(this)()`)
};

editor.DomComponents.addType(compType, {
    model: {
      defaults: {
        script,
        userscript: userScript1,
        'script-props': ['userscript'],
        traits: [
          {
            type: 'button',
            name: 'edit user script',
            text: 'Edit user script...',
            full: true,
            command: 'edit-script'
          }
        ]
      },
```

A huge positive is that because the component scripts aren't touched by users, the component script remains a function object and thus keeps updating itself as the app / component implementation script change over time. No component update problems anymore.

The main downside is that only components that I create myself, with those two magic lines of code in their official component script,  can support user scripts.  That's probably OK for my use case. 

I'll continue to think about how this might be integrated deeper into Grapesjs so that *any* arbitrary component can have a 'user' script', which would be kind of cool.

PR #3308, opened 2021-03-04
assign target in trait if its added in a target format

jcamejo:
Hi @artf, I hope everything's good.

When I was trying to add a select trait dynamically as an object on a component, I've noticed that **getInputEl** of the **traitSelectView** tried to access the target with **model.getTargetValue()** and the target was not initialized yet, the target is assigned in the collection only if you pass the trait as an array or a string.

Let me know if i have to validate something else or add this action in another file. 

Thanks!

### DIFF
--- a/src/trait_manager/model/Traits.js
+++ b/src/trait_manager/model/Traits.js
@@ -1,5 +1,5 @@
 import Backbone from 'backbone';
-import { isString, isArray } from 'underscore';
+import { isString, isArray, isObject } from 'underscore';
 import Trait from './Trait';
 import TraitFactory from './TraitFactory';
 
@@ -30,6 +30,10 @@ export default Backbone.Collection.extend({
 
   add(models, opt) {
     const em = this.em;
+
+    if (isObject(models)) {
+      models.target = this.target;
+    }
 
     // Use TraitFactory if necessary
     if (isString(models) || isArray(models)) {


artf:
Hi @jcamejo thanks for the PR and sorry for the late reply. You know, I've tried to add dynamically the trait (of select type), but didn't face any issue or error. Are you able to provide the necessary steps for the reproduction?

jcamejo:
Hi @artf.

I am also sorry for the late reply, this merge is not necessary, it was a mistake for not using the right API on the component

I was doing this in the initialization of a custom component

  ```
    var method = {
        type: 'select',
        name: 'method',
        options: [
          { value: '1', name: '1'},
          { value: '2', name: '2'}
        ],
        changeProp: 1
      };

      model.addTrait(method);

```

After some other initialization instructions I've added more traits in an array that included that same **method** object but with a big difference. Instead of using **addTrait** , the traits collection was being accessed directly. Doing so i was able to remove all the traits from the component but it also bypasses the **add** function of the trait collection that is not longer triggered, prokoving the **target** undefined issue as well.

`      model.get('traits').set([trait1, trait2, method]);`

Do you have any suggestions on how to remove all the traits of a component? I am thinking on doing this

 model.get('traits').set([]);
 model.get('traits').addTrait([..traits..]);

You can close this merge if you don't see anything unusual.

Thanks!

artf:
Well yeah you can use addTrait/removeTrait, but I'd expect this to work `model.set('traits', [trait1, trait2, method]);`, as the component should recreate the collection on its change

jcamejo:
I will see what i can do!

artf:
Hey @jcamejo do you think we can close this one?

jcamejo:
I have not checked what was happening in the set method of the traits collection. But in relation to the real purpose of the issue, we can close it.

artf:
Great, thanks Juan

PR #3314, opened 2021-03-05
Support image srcset attribute

aimeos:
Adds image srcset support for responsive images added to the asset manager using:
```js
editor.AssetManager.add({
    src: 'http://path/to/image.jpg',
    srcset: 'http://path/to/image-200.jpg 200w, http://path/to/image-800.jpg 800w'
});
```

### DIFF
--- a/src/asset_manager/model/AssetImage.js
+++ b/src/asset_manager/model/AssetImage.js
@@ -5,6 +5,7 @@ export default Asset.extend({
     ...Asset.prototype.defaults,
     type: 'image',
     unitDim: 'px',
+    srcset: '',
     height: 0,
     width: 0
   }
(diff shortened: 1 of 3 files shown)


artf:
Hi @aimeos thanks for the PR. The usage of `srcset` is cool, but IMHO it's more an application-level specification and not of the framework itself, otherwise, we would need to keep extending components and assets for each use case, and it's not a scalable solution for a framework. 
For your case, I'd only suggest extending the Image component type, in order to support the `srcset`. I guess something like this should work
```js
editor.Components.addType('image', {
 view: {
   onActive(ev) {
     ev && ev.stopPropagation();
     const { model } = this;
     editor.runCommand('core:open-assets', {
        target: model,
        types: ['image'],
        onSelect(asset) {
          const srcset = asset.get('srcset');
          srcset && model.addAttributes({ srcset })
        }
      });
   }
 }
})
```

PR #3318, opened 2021-03-05
Set width of side panel to 20% (Fixes #3312)

### LINKED ISSUE #3312: FEAT: Make side panel's width to 20%
**What are you trying to add to GrapesJS?**

The side panel looks scrunched. Unless you've got a big desktop, I'd rather not have images from the blocks panel go off to the side. Additionally, the icons at the top e.g. "getting a new block" icon is just obscured unless you've got a decently large screen.

*Describe your feature request **detailed***

Changing the width of the side panel from 15% to 20%, specifically the class: `gjs-pn-views-container`

**Is there an alternative at the latest version?**

[ ] Yes (descripe the alternative)
[X] No

**Is this related to an issue?**

[ ] Yes (Give a link to the issue)
[X] No

Andrew-Chen-Wang:
Sets the width of the side panel to 20% so the panel with all the blocks doesn't look scrunched up on a medium sized (browser) screen.

I'm also thinking about when all three side panel options are not selected, the side panel is hidden. Kinda of like a preview view but you can still drag the blocks.

### DIFF
--- a/src/styles/scss/_gjs_variables.scss
+++ b/src/styles/scss/_gjs_variables.scss
@@ -41,7 +41,7 @@ $colorGreen:        #62c462 !default;
 $tagBg:             #804f7b !default;
 $secColor:          $tagBg !default;
 $imageCompDim:      50px !default;
-$leftWidth:         15% !default;
+$leftWidth:         20% !default;
 
 /* Color Helpers */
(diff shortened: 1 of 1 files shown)


Andrew-Chen-Wang:
Huh interestingly the canvas is now gone...

PR #3338, opened 2021-03-14
Transform  Backbone extend to es6 class extends

cereschen:
In this way, with @types/backbone, we can get a good code editor intellisense.


artf:
Well actually this is good, also for the debugging purpose, but I really don't like the `CLASS.prototype.PROP` 😬
Probably we have to add [this](https://babeljs.io/docs/en/babel-plugin-proposal-class-properties) to grapesj-cli to solve that

artf:
Anyway, I'm planning to merge soon a big update (support for multi-pages), probably it makes sense to add such a change after the merge, in order to avoid all the conflicts.

cereschen:
After that, I will open the pr again.

PR #3360, opened 2021-03-23
Adds a title attribute to layers panel elements

### LINKED ISSUE #3355: FEAT: Add a title attribute to the Layers panel elements
Hi @artf ,

When I have several elements nested in the Layers panel then some elements names are truncated because the Layer panel's width is not big enough, I know that custom resizable panel can be a solution but adding a title attribute would be enough for the majority of the use cases.

Screenshot 2021-03-22 at 14 50 53

A tooltip (title) would be helpful to the user.

**grapesjs/src/navigator/view/ItemView.js**
May be adding a title attribute to the follow line can do the job?
https://github.com/artf/grapesjs/blob/7a23e20fe5138c37088ef05382f7c759cb6fa951/src/navigator/view/ItemView.js#L47  
  
Example:
```javascript
<div class="${pfx}layer-title-inn" title="${name}">
```

Let me know what do you think,

Thanks!

Franci5co5aoco:
This PR adds a title attribute to layers panel elements.

Closes #3355

### DIFF
--- a/src/navigator/view/ItemView.js
+++ b/src/navigator/view/ItemView.js
@@ -47,7 +47,7 @@ export default Backbone.View.extend({
       }
       <div class="${clsTitleC}">
         <div class="${clsTitle}" style="padding-left: ${gut}" data-toggle-select>
-          <div class="${pfx}layer-title-inn">
+          <div class="${pfx}layer-title-inn" title="${name}">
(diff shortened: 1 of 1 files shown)


artf (approved):

artf:
Great, thanks 👍

PR #3361, opened 2021-03-23
Feature/media condition per device

artenc:
Sometimes there is a need to specify media condition for every device separately, for example, if need a two direction style cascading as it's in Webflow - when there is middle breakpoint ("desktop") and rules are applied in direction of smaller devices or larger desktops.

So now each device config also supports "mediaCondition" parameter. It overrides global "mediaCondition" setting if specified.

I made changes cross compatible so that original configuration works the same way. However I also fixed the default device config for Tablet, which was overlapping with Mobile Landscape widthMedia.


src/device_manager/model/Device.js
```diff
@@ -16,6 +16,10 @@ export default Backbone.Model.extend({
     // If empty the width will be used
     widthMedia: null,
 
+    // The condition to use for media queries, eg. 'max-width'
+    mediaCondition: null,
+    defaultMediaCondition: null,
```
artf, line 21: I think there is no need for defaultMediaCondition, you can get it directly from the config

src/css_composer/view/CssRulesView.js
```diff
@@ -104,10 +104,7 @@ export default Backbone.View.extend({
 
   getMediaWidth(mediaText) {
     return (
-      mediaText &&
-      mediaText
-        .replace(`(${this.em.getConfig('mediaCondition')}: `, '')
-        .replace(')', '')
+      mediaText && mediaText.replace(/\((min|max)-width: /, '').replace(')', '')
```
artf, line 107: Well, I know that min/max-width are the most used conditions but what if I want to use something completely custom. I'd leave your new one, but I'd restore the replace with mediaCondition from the config

src/device_manager/model/Devices.js
```diff
@@ -5,8 +5,20 @@ export default Backbone.Collection.extend({
...
+
+    if (!leftC || !rightC) {
+      return leftC == rightC || !rightC ? 1 : -1;
+    }
+
+    if (leftC == rightC) {
+      return (rightP - leftP) * (leftC == 'max-width' ? 1 : -1);
+    }
+
+    return rightC == 'max-width' ? -1 : 1;
```
artf, line 21: IMHO it's a bit confusing logic here, can you add some comments to explain what is actually happening here? Probably adding some tests would be even better

artf (changes requested):

artf:
Hey @artenc any update on this one?

artf:
@artenc I'm closing this for inactivity. If you want you can submit a new one with my suggestions.

PR #3391, opened 2021-04-06
Update _gjs_canvas.scss, allow vars to override and fix missing use of $canvasTop

nnmer:
Update _gjs_canvas.scss, allow vars to override and fix missing use of $canvasTop


artf (approved):

artf:
Thanks Anton 👍

Issue #3397, opened 2021-04-09
BUG: sorter direction incorrect with web component slot


YumiChen:
Version:0.16.45

Are you able to reproduce the bug from the demo?

 - [x] Yes
 - [ ] No

What is the expected behavior?
Given a parent web component instance rendering some elements with `slot`, in the shadowDOM, if the parent element of `slot` got `display: flex` style, the sorter should be vertical, for in the result the elements are aligned horizontally.

**What is the current behavior?**
The sorter would be horizontal.

**Are you able to attach screenshots, screencasts or a live demo?**

 - [x] Yes (attach)
 - [ ] No

https://jsfiddle.net/72dyhu6L/
<img width="741" alt="截圖 2021-04-09 下午6 38 06" src="https://user-images.githubusercontent.com/27608365/114168501-e56e7c00-9962-11eb-9b0b-deaf86ba5e73.png">

Issue #3396, opened 2021-04-09
BUG: flex-direction: column & float: left make sorter direction incorrect


YumiChen:
Version:0.16.45

Are you able to reproduce the bug from the demo?

 - [x] Yes
 - [ ] No

What is the expected behavior?
Given a parent element with `display: flex` and `flex-direction: column` styles, with the child elements having `float: left` styles, the sorter should be horizontal, for in the result the elements are aligned vertically.

**What is the current behavior?**
The sorter would be vertical.

**Are you able to attach screenshots, screencasts or a live demo?**

 - [x] Yes (attach)
 - [ ] No

https://jsfiddle.net/fzb4u7rd/
<img width="741" alt="截圖 2021-04-09 下午6 18 38" src="https://user-images.githubusercontent.com/27608365/114166272-1bf6c780-9960-11eb-91d0-ceb1454d81f4.png">

PR #3398, opened 2021-04-09
Fix sorter direction

YumiChen:
Hi @artf, thanks very much for the good work. 

I was trying to integrate web components into GrapesJS and bumped into some sorter direction issues while sorting the components( Related issues are raised: #3396 & #3397) Changes here are added to handle the scenarios mentioned in the issues.

If there's anything missing in the changes please let me know. Thank you!

### DIFF
--- a/src/utils/Sorter.js
+++ b/src/utils/Sorter.js
@@ -565,6 +565,14 @@ export default Backbone.View.extend({
       $parent.css('flex-direction') !== 'column'
     )
       return;
+    if (
+      $parent &&
+      $parent.css('display') == 'flex' &&
+      $parent.css('flex-direction') === 'column'
+    )
+      return true;
+    if ($el.css('float') !== 'none') return;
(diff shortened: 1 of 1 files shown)


src/utils/Sorter.js
```diff
@@ -555,16 +555,20 @@ export default Backbone.View.extend({
     if (isTextNode(el)) return;
     const style = el.style || {};
     const $el = $(el);
-    const $parent = parent && $(parent);
+    let $parent = parent && $(parent);
+    if(parent.shadowRoot){
+      const slot = parent.shadowRoot.querySelector('slot');
+      if(slot) $parent = $(slot).css('display') === 'contents'?$(slot.parentElement):$(slot);
```
artf, line 561: Are you able to illustrate with an example such a case? I mean, is it valid for general use of Web Components or it fixes only your case?

PR #3436, opened 2021-05-05
Added Functionality to Call the Parent on the Resizer Functions.

### LINKED ISSUE #3435: BUG: Scrollable onStart/onEnd/updateTarget not Overridable
**Version:** 0.17.3

**Are you able to reproduce the bug from the demo?**

 - [ ] Yes
 - [X] No

Being able to override onStart, onEnd, and updateTarget in a component's `model.default.scrollable`, as detailed in [the documentation](https://github.com/artf/grapesjs/blob/master/src/utils/Resizer.js).

I have created a component that extends ComponentImage. The content is not a static image but rendered as necessary, and when the destination size changes, it has to be rerendered.

According to the documentation, I should be able to define `onEnd` in `model.default.scrollable` to trigger the rerender from there after the resizing is done (I don't want to continually redraw while the user is still resizing).

When I define any `onEnd` function, it overrides the `onEnd` defined [in SelectComponent.js](https://github.com/artf/grapesjs/blob/210f87030d7d1e2aa2e5c427e4347b49ad258134/src/commands/view/SelectComponent.js#L453), which breaks the whole state machine there. For example, since `toggleFramesEvents` is no longer called, the whole iframe is wrapped in `pointer-events: none` and thus doesn't respond to any user interaction any more.

Since this is not a class I'm overriding, but just some inline functions that get overwritten, I can't call the parent function, since it doesn't exist any more within the scrollable. I also can't copy/paste the content of the parent function into mine, because it references local variables I don't have access to.

The same issue exists with all other functions that are defined in SelectComponent.js there, which are onStart, onMove, and updateTarget.

anlumo:
Fixes #3435.

### DIFF
--- a/src/commands/view/SelectComponent.js
+++ b/src/commands/view/SelectComponent.js
@@ -494,7 +494,7 @@ export default {
       };
 
       if (typeof resizable == 'object') {
-        options = { ...options, ...resizable };
+        options = { ...options, ...resizable, parent: options };
       }
 
       this.resizer = editor.runCommand('resize', { el, options, force: 1 });


artf (approved):

artf:
Nice workaround 👍

PR #3456, opened 2021-05-12
Edit-only selectors feature

artenc:
Sometimes I need to mark specific selectors as edit-only, which means that they should not be rendered for output. More specifically, if selector editonly attribute is true, then Css/Html code generators will not include the rule associated with the selector.

### DIFF
--- a/src/code_manager/model/CssGenerator.js
+++ b/src/code_manager/model/CssGenerator.js
@@ -107,19 +107,22 @@ export default Backbone.Model.extend({
     const selectorStrNoAdd = rule.selectorsToString({ skipAdd: 1 });
     const selectorsAdd = rule.get('selectorsAdd');
     const singleAtRule = rule.get('singleAtRule');
-    let found;
 
-    // This will not render a rule if there is no its component
-    rule.get('selectors').each(selector => {
-      const name = selector.getFullName();
-      if (
-        this.compCls.indexOf(name) >= 0 ||
-        this.ids.indexOf(name) >= 0 ||
-        opts.keepUnusedStyles
-      ) {
-        found = 1;
-      }
-    });
+    const selectors = rule.get('selectors').models;
+
+    // This rule will render in case:
+    // * all of selectors are not editonly
+    // * if some component is using the rule or keepUnusedStyles is enabled
+    const found =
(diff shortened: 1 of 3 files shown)


artf:
@artenc can you illustrate a case example where I would need to have selectors which are not rendered in the output? I don't understand where is the point.

artenc:
Hi @artf, sure, let me try to explain.

There are cases, when editing a page, we need to decorate a component in some way (e.g. adding padding, so that particular component can be easier selected).
This can be done of course on view.onRender, by adding styles or classes to component without changing a model itself. There is a problem though, when some attribute (including style or class) changes it's value through model, we loose our styles set from view.onRender.
Of course there is a workaround, we can listen to attributes changes and adding our editonly styles again, but it's just not comfortable for us to make this kind of logic for each component.

Instead we do it through selectors.
In our project we extend functionality in a way, so that instead of specifying classes in component model, we can specify selectors and their css rules. These selectors and rules associated with component, are then created (if they don't exist) using SelectorManager and CssComposer.
So we have ability to mark some of selectors as editonly, and eliminating need to define additional repeating logic in component view.

Hope this answers your question.

artf:
> There is a problem though, when some attribute (including style or class) changes it's value through model, we loose our styles set from view.onRender.

Well, you should not rely on styles and classes :) GrapesJS appends always `[data-gjs-type=NAME]` to components in canvas, so, create a simple helper which appends styles related to that component

```js
function addStyle(view, type, customStyles) {
  const key = 'data-type';
  const keyAttr = `[${key}="${type}"]`;  
  const head = view._getFrame().getHead();  
  if (!head.querySelector(`style${keyAttr}`)) {
    const style = document.createElement('style');    
    style.setAttribute(key, type);
    style.innerHTML = `[data-gjs-type="${type}"] { 
      ... print here your customStyles
     }`;
    head.appendChild(style);
  }
}
editor.Components.addType('mycomp', {
 model: { ... },    
 view: {      
  onRender() {
    addStyle(this, 'mycomp', { padding: '10px' });      
  },    
 }  
});
```

artf:
Hey @artenc I'm closing this one as I believe such a feature is not necessary (check my example above)

PR #3505, opened 2021-05-31
Typo Fixed

meamitacharya:


artf (changes requested):
This file is generated from https://github.com/artf/grapesjs/blob/dev/src/asset_manager/index.js :)

artf:
Updated in dev

PR #3550, opened 2021-06-18
Only patch keyboard events when creating a custom event.

### LINKED ISSUE #3549: BUG: createCustomEvent Doesn't Handle Mouse Events Correctly
**Version:** 0.17.4 / git master

FrameView repackages keyboard and mouse events here:

https://github.com/artf/grapesjs/blob/07061ae1307fedf2a3b9bd585ce0cb30f653f112/src/canvas/view/FrameView.js#L389-L399

I think the reason is related to redirecting iframe events to parent frame events. This would be fine, however, the `createCustomEvent` function looks like this:

https://github.com/artf/grapesjs/blob/07061ae1307fedf2a3b9bd585ce0cb30f653f112/src/utils/dom.js#L63-L81

As you can see, the `which` property is defined to return the value of `keyCode`. This works for keyboard events, but on mouse events, it should return the value of `which` from the wrapped event.

This causes `which` to be `undefined` in the event, breaking an external package I use that relies on `event.which` for mouse events.

anlumo:
Fixes #3549


artf (approved):

artf:
Thanks @anlumo I apologize for the late reply on this. 😁

PR #3570, opened 2021-06-26
Fixed typo

SingingApple:
Change **catgegory** to **category** on line 52.


artf:
Hi @SingingApple that file is autogenerated from [here](https://github.com/artf/grapesjs/blob/dev/src/block_manager/index.js), so you should update the source in case of API references. I'll update the source now, thank you.

PR #3628, opened 2021-07-18
Layer manager width now scales with content

MortenSpjotvoll:
Currently when you go in depth in the layer manager the scaling gets completely messed up like this:
https://i.imgur.com/gE20I9G.png

By simply adding display:grid to the .gjs-layer we instead get a scalable solution that always fits all the children and shows all the content
https://i.imgur.com/I7zfz4m.png


artf:
Hi @MortenSpjotvoll the solution looks great, can you apply the change here https://github.com/artf/grapesjs/blob/dev/src/styles/scss/_gjs_layers.scss and leave the build file out of the PR, thanks

MortenSpjotvoll:
> Hi @MortenSpjotvoll the solution looks great, can you apply the change here https://github.com/artf/grapesjs/blob/dev/src/styles/scss/_gjs_layers.scss and leave the build file out of the PR, thanks

Thanks @artf . That makes a lot more sense indeed. I have corrected it now

artf (approved):

artf:
Great, thanks Morten 👍

PR #3828, opened 2021-09-27
Fix left position of rteToolbarPosUpdate

bernhardmiller:
With this patch, the RTE position as modified by "rteToolbarPosUpdate" (see for example https://github.com/artf/grapesjs-plugin-ckeditor/blob/master/src/index.js#L121-L142) is now applied instead of thrown away.


artf (approved):

artf:
Thanks @bernhardmiller

PR #3942, opened 2021-11-14
Fix unexpected trait type change in TraitManager.addType() call

xQwexx:


artf:
Hey @xQwexx can you elaborate on this change? This basically breaks how new Trait types are [created](https://grapesjs.com/docs/modules/Traits.html#define-new-trait-type)

xQwexx:
OK I understand this is the naming konvention for this framework, but it is really bad name for this type of functions because from a programming perfective I would expect the addType() to want a type class not a config attr., because of this naming I spent few hours to debug... So I would suggest renaming this function something like defineType() or defineNewType() which would suggest that the type definition happens here so its more intuitive

artf:
How `defineType` is different from `addType`, from your programming perspective?
I might agree with you in case I'd call it `addClassType`, but this is not the case and the documentation is quite clear about the definition.
I have to close it as it's a useless breaking change.

I'd recommend opening a discussion before creating a PR containing a breaking change. That's also to avoid wasting your time on creating something that won't be accepted.

PR #3952, opened 2021-11-17
Adjust draggable property to accept function

gnbrndn:
Hi @artf,

Inline with #3393. I've adjusted `draggable` to accept function. It will receive the source and target models as argument and needs to return a boolean value.


artf (approved):

artf:
Thanks @gnbrndn makes sense 👍

PR #3992, opened 2021-12-03
Allow for the title of a block to be passed as an attribute

koendeschacht:
The documentation seems to suggest that this should already be possible, but currently the title passed as an attribute is ignored.

https://grapesjs.com/docs/api/block_manager.html#add


artf (approved):

artf:
Good catch, thanks @koendeschacht

Issue #3589, opened 2021-07-03
Autoload storage is not working on page refresh


lbmir:
Hi, firstly I want to say that this project is awesome, thank you!

I used code from GrapeJS Demo example, and added there save and load data from remote files, this is my storageManager code:

```
 storageManager: {
            id: 'gjs-',             // Prefix identifier that will be used on parameters
            type: 'remote',          // Type of the storage
            autosave: false,         // Store data automatically
            autoload: true,         // Autoload stored data on init
            urlStore: 'save_now.php',
            urlLoad: 'load_now.php',
            params: { page_id: 11111 },
            contentTypeJson: true,
            headers: {
            'Content-Type': 'application/json'
            }
        }
```

save_now.php file code:

```
header("Content-Type:application/json");
$data = file_get_contents("php://input");
file_put_contents('test.txt', $data);
```

load_now.php file code:

```
header("Content-Type:application/json");
$response = file_get_contents("test.txt");
echo $response;
```

When I am saving my changes it is working fine, but after that if I am press to **Refresh** page it loads old content which was before my changes.

But if I use **Hard Refresh** page than it is content is fine it is showing my last changes.

How can I keep saved content when user refreshing page?

It seems that it is saving old content somewhere and reloading it after refresh but not using urlLoad from storage manager that I defined. 

May be there is a way to disable loading of old content and only use urlLoad from storage Manager?

lbmir:
Same issue like on [ https://github.com/artf/grapesjs/issues/2963]( https://github.com/artf/grapesjs/issues/2963).
But in my case it is loading content after Hard Refresh

artf:
Sounds more like an issue from your side, web applications do not distinguish Refresh from Hard Refresh, indeed the Hard Refresh is more like a refresh without the browser local cache. Try to check, from the inspector, if your load call is actually returning the correct data (eg. maybe you have a web server with kind of cache enabled)

PR #4049, opened 2022-01-06
Fix race condition in CssComposer loading.

bernhardmiller:
At the time the load() function is called, the CssComposer could already have some rules added to it.

When loading a grapes definition with inline styles, it is possible that some of the inline-defined CSS rules are overwritten by the CssComposer load. This is because CssComposer->load reset the rules on load.

The race is between the components loader already adding new rules to CssComposer before CssComposer->load is being called.

I could only reliably reproduce the bug on slow internet connections (using Chrome Dev Tools, I had to simulate a slow 3G connection to make it happen reliably). It is also important that some of the components being loaded the earliest have some inline style to them, like this:
```
[
  {
    "type": "my-wrapper",
    "components": [
      // ...
    ],
    "style": {
      "__bg-type": "color",
      "background-image": "linear-gradient(#FF0000,#FF0000)",
      "background-repeat": "repeat",
      "background-position": "left top",
      "background-attachment": "scroll",
      "background-size": "auto"
    }
  }
]

```

The order of CssComposer function calls is sometimes like this, which produces a good result:
 * load()
 * setIdRule
 * setIdRule
 * // ...

But sometimes it is also like this, which leads to the first two rules being deleted:
  * setIdRule
  * setIdRule
  * load
  * setIdRule
  * // ...

The reason I'm explaining this in so much detail is that I don't know if the same problem can also be found in other parts of grapes. I am not sure what side effects (other than adding css rules) the loading of components might have; which are later overwritten by forceful resetting in the load function.
For example, #3589 could be related. While investigating this problem, the loading *always* worked correctly with a hard refresh, but failed relatively often with a normal refresh.

### DIFF
--- a/src/css_composer/index.js
+++ b/src/css_composer/index.js
@@ -134,9 +134,9 @@ export default () => {
       }
 
       if (isArray(obj)) {
-        obj.length && rules.reset(obj);
+        obj.length && rules.add(obj);
       } else if (obj) {
-        rules.reset(obj);
+        rules.add(obj);
       }
 
       return obj;


bernhardmiller:
I just noticed that this fix is probably good for my use-case; but will not work in general. Other people might call StorageManager->load multiple times, which would break their documents.

This will probably need an additional preLoad function (similar to postLoad) that cleans the current state before calling the fixed load function.

artf:
Hi @bernhardmiller, yeah, adding rules instead of resetting them in `load` will not work with multiple Storage load calls so, it's not an option.
But I can actually imagine this issue happening by checking this old snippet
https://github.com/artf/grapesjs/blob/b2d0ac55653b5d6847bf5f24ddeb1445575ccccc/src/editor/model/Editor.js#L220-L225
Honestly, I can't recall why I prepend the Components module here but this actually can trigger the issue you're describing. 
I can probably guess that this snippet doesn't make sense anymore (not even from the logical point of view) and the condition for `domComponents` could be removed.
Can you check please if you're able to reproduce the issue without that condition?

artf:
@bernhardmiller I'm closing this one, for now, as it can't be applied but let me know if the alternative fix works for you.

bernhardmiller:
@artf Sorry for being ultra slow to respond, this one almost fell through the cracks.
I tried the fix you suggested, and cannot reproduce the error if I change the code as you suggested:

```
 if (Mod.storageKey && Mod.store && Mod.load && sm) { 
   cfg.stm = sm; 
   this.get('storables').push(Mod); 
 } 
```

Before I open a new merge request with this fix; can you please let me know if the order of the storables is defined somewhere? It was consistent through my experiments, but if some random occurrence (or future code change?) makes domComponents come before cssComposer again, the error will hit again.

So I'm not sure if the above fix is enough, or if it's better to explicitely put cssComposer in front, similar to how it's done right now for domComponents:

```
 if (Mod.storageKey && Mod.store && Mod.load && sm) { 
   cfg.stm = sm; 
   // CSS Composer should be load before DOM Components
   const mth = name == 'cssComposer' ? 'unshift' : 'push'; 
   this.get('storables')[mth](Mod); 
 } 
```

artf:
@bernhardmiller I'm actually working on the StorageManager refactoring and as I faced similar issues I'll address them there. Unfortunately, I've realized that the current order makes sense in some cases so I'd like to remove this unstable order dependency.
So I'd ask to wait for the `v0.19.*` release.

PR #4103, opened 2022-01-26
Document component->style

lofcz:
@artf shouldn't style be defined as `{}` by default?


artf:
Thanks @lofcz, yeah you're right, but there are some issues related to default array/objects with how I've started defining default components so I'll need to deep check some stuff in order to avoid regression.
For now, I'm ok with having `[style={}]` in docs and an empty string in code 👍

lofcz:
Updated

artf (approved):

PR #4128, opened 2022-02-05
Up docs

lofcz:
document `getCss->opts:keepUnusedStyles`


artf (changes requested):
Thanks @lofcz but files in `/api` folder are actually generated directly from the source files, so you should update this one https://github.com/artf/grapesjs/blob/dev/src/editor/index.js

lofcz:
Sorry, fixed

artf (approved):

PR #4140, opened 2022-02-13
Fix empty scrips gen

lofcz:
Components with an empty script in definition produce unnecessary js in output.
A component defined like:
```js
model: ...
 defaults {
  script: () => {}
 }
```
When used and scripts are exported this results in something like:
image

We can skip this in cases of scripts's content being either whitespace only or a falsy value.

### DIFF
--- a/src/code_manager/model/JsGenerator.js
+++ b/src/code_manager/model/JsGenerator.js
@@ -39,7 +39,7 @@ export default Backbone.Model.extend({
       }
     }
 
-    comps.each(function(model) {
+    comps.each(function (model) {
       code += this.mapModel(model);
     }, this);
 
@@ -55,7 +55,8 @@ export default Backbone.Model.extend({
       const mapType = this.mapJs[type];
 
       if (mapType.props) {
-        code += `
+        if (mapType.code && mapType.code.trim()) {
+          code += `
           var props = ${JSON.stringify(mapType.props)};
           var ids = Object.keys(props).map(function(id) { return '#'+id }).join(',');
           var els = document.querySelectorAll(ids);
@@ -63,17 +64,20 @@ export default Backbone.Model.extend({
             var el = els[i];
             (${mapType.code}.bind(el))(props[el.id]);
           }`;
+        }
       } else {
         // Deprecated
-        const ids = '#' + mapType.ids.join(', #');
-        code += `
+        if (mapType.code && mapType.code.trim()) {
+          const ids = '#' + mapType.ids.join(', #');
+          code += `
           var items = document.querySelectorAll('${ids}');
           for (var i = 0, len = items.length; i < len; i++) {
             (function(){\n${mapType.code}\n}.bind(items[i]))();
           }`;
+        }
       }
     }
 
     return code;
-  }
+  },
 });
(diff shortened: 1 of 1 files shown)


src/code_manager/model/JsGenerator.js
```diff
@@ -55,25 +55,29 @@ export default Backbone.Model.extend({
...
           for (var i = 0, len = els.length; i < len; i++) {
             var el = els[i];
             (${mapType.code}.bind(el))(props[el.id]);
           }`;
+        }
       } else {
         // Deprecated
-        const ids = '#' + mapType.ids.join(', #');
-        code += `
+        if (mapType.code && mapType.code.trim()) {
```
artf, line 70: It's quite safe to assume that `mapType.code` will always be defined and trimmed (from `model.getScriptString`).
In this case, I also think we can avoid duplicating the same logic by simply checking on top
```js
...
if (!mapType.code) continue;

if (mapType.props) {
...
```

artf (changes requested):
Overall the improvement is good and makes totally sense, just check the comment below.

I'd also like to point out that right now we're using 2 different code generation approaches.
The old one, for these cases (which might be removed in the future):
```js
defaults: {
  //...
 script: () => {/* script content */}
}
```
and the new one if `script-props` is defined:
```js
defaults: {
  //...
 'script-props': ['prop1', 'prop2'],
 script: (props) => {/* script content */}
}
```
By checking `Component.getScriptString` you will realize that the script string for the first case (deprecated) is extracted from the function string `function() { [content here is extracted] }` but in the second case, we just stringify the function `script.toString().trim()`, so the initial string `function(...` will always be here and the proposed fix won't never work for the new approach.
If we want to support also the new script generation approach we need a new check inside the `if (mapType.props) {` condition.
```js
if (mapType.props) {
 if (isFunctionEmpty(mapType.code)) continue;
 ...
}
```

src/code_manager/model/JsGenerator.js
lofcz: Thanks for taking time to explain these concepts, much appreciated. I have a question regarding implementation of `isFunctionEmpty`. I've never needed this before and it seems to be a [complicated topic](https://stackoverflow.com/questions/7960335/javascript-is-given-function-empty).

After the first assert (`if (!mapType.code) continue;`) would asserting `isFunctionEmpty(mapType.code)` via this be safe?
```js
function isFunction(fn) {
  const content = fn.toString().match(/\{([\s\S]*)\}/m)[1]; // content between first and last { }
  return content.replace(/^\s*\/\/.*$/mg,'').length > 0; // remove comments
}
```
Can't we do something like minify the script with `uglify` and then check the length? That would also get us rid of adding newlines everytime in the old approach since https://github.com/artf/grapesjs/pull/3300 due to a rather edge-case scripts containing comments on the first line. Compressing emitted js would a great feature but it might be out of scope of this PR.

artf: The `isFunctionEmpty(mapType.code)` should be added inside `if (mapType.props) {` as the old approach tries already to clean the function signature.
In all honesty, I wouldn't spend too much time and effort on that check. Your `isFunctionEmpty` example is already cool and more than enough, I wouldn't even care about removing comments.
The minifying option is for sure a nice thing to have but I always thought about them as an extension/plugin for code generators but I still need to refactor and improve the API of this module to support such functionality.

artf (approved):

artf:
Well done @lofcz Thank you very much

PR #4149, opened 2022-02-18
Update CONTRIBUTING.md

Joel286:


Joel286:
First tryout

artf:
@Joel286 what were you trying to accomplish here? I only see that you're breaking the Open Collective link

PR #4218, opened 2022-03-28
Fix terser-webpack-plugin not found

xQwexx:
When I cloned the repo I run the npm install and tried to build the project but it failed with terser-webpack-plugin not found so I added it to the dev-dependecies.


artf (approved):
I'm not sure why the terser plugin is not found in grapesjs-cli but I guess it makes sense from the script point of view. I'll add this fix for now and will refactor the next version of the cli in order to remove the need for this dependency.

PR #4219, opened 2022-03-28
Use class definitions instead of default object

xQwexx:


artf (changes requested):
I'm seeing this on the dev server
Schermata 2022-03-29 alle 16 01 15
Schermata 2022-03-29 alle 16 01 25

I'm not sure the class properties you refactored would work as expected. From what I know es6 class methods are merged on the prototype but not properties.
https://stackoverflow.com/questions/38311805/cannot-define-prototype-properties-within-es6-class-definition

This breaks some of the backbone extend mechanisms.

You can actually see from other places where I use classes with backbone models/views I'm adding them manually to the prototype.

xQwexx:
Sorry I forget to run the test I just used the browser and I didn't errors there so I thought everything ok, I updated with the changes.

artf (approved):
Awesome @xQwexx thanks you very much

PR #4220, opened 2022-03-28
Fix  BUG: Dragging multiple blocks reverses the order #3547

### LINKED ISSUE #3547: BUG: Dragging multiple blocks reverses the order
### Version:

0.17.4

### Are you able to reproduce the bug from the demo?

YES

### What is the expected behavior?

When dragging the blocks, the order should be preserved.

### **What is the current behavior?**

When dragging to the top or above the set of blocks, it goes in one order.

When dragging to the bottom or below the set of blocks, it goes in another order.

This may become a problem if selecting a lot of blocks or the blocks are tall.

### **Are you able to attach screenshots, screencasts or a live demo?**

Screen Shot 2021-06-17 at 10 34 58 AM

Screen Shot 2021-06-17 at 10 35 04 AM

https://user-images.githubusercontent.com/676680/122304455-5f217800-cf59-11eb-8925-0374deb9b1fb.mov

lexoyo:
Hello
I'm trying to fix #3547 
This is more of an exercise for me to dive in grapesjs code, let me know if I'm loosing my time or if it sounds useful

### DIFF
--- a/src/utils/Sorter.js
+++ b/src/utils/Sorter.js
@@ -983,6 +983,35 @@ export default Backbone.View.extend({
   },
 
   /**
+   * Build an array of all the parents, including the component itself
+   * @return {Model|null}
+   */
+  parents(model) {
+    return model ? [model].concat(this.parents(model.parent())) : [];
+  },
+
+  /**
+   * Sort according to the position in the dom
+   * @param {Object} obj1 contains {model, parents}
+   * @param {Object} obj2 contains {model, parents}
+   */
+  sort(obj1, obj2) {
+    // common ancesters
+    const ancesters = obj1.parents.filter(p => obj2.parents.includes(p));
+    const ancester = ancesters[0];
+    if (!ancester) {
+      // this is never supposed to happen
+      return obj2.model.index() - obj1.model.index();
+    }
+    // find siblings in the common ancester
+    // the sibling is the element inside the ancester
+    const s1 = obj1.parents[obj1.parents.indexOf(ancester) - 1];
+    const s2 = obj2.parents[obj2.parents.indexOf(ancester) - 1];
+    // order according to the position in the DOM
+    return s2.index() - s1.index();
+  },
+
+  /**
    * Leave item
    * @param event
    *
@@ -1013,9 +1042,43 @@ export default Backbone.View.extend({
     if (this.moved && target) {
       const toMove = this.toMove;
       const toMoveArr = isArray(toMove) ? toMove : toMove ? [toMove] : [src];
-      toMoveArr.forEach(model => {
-        moved.push(this.move(target, model, lastPos));
-      });
+      let domPositionOffset = 0;
+      if (toMoveArr.length === 1) {
+        // do not sort the array in this case
+        // there are cases for the sorter where toMoveArr is [undefined]
+        // which allows the drop from blocks, native D&D and sort of layers in Style Manager
+        this.move(target, toMoveArr[0], lastPos);
+      } else {
+        toMoveArr
+          // add the model's parents
+          .map(model => ({
+            model,
+            parents: this.parents(model),
+          }))
+          // sort based on elements positions in the dom
+          .sort(this.sort)
+          // move each component to the new parent and position
+          .forEach(({ model }) => {
+            // store state before move
+            const index = model.index();
+            const parent = model.parent().getEl();
+            // move the component to the desired position
+            moved.push(
+              this.move(target, model, {
+                ...lastPos,
+                indexEl: lastPos.indexEl - domPositionOffset,
+                index: lastPos.index - domPositionOffset,
+              })
+            );
+            // when the element is dragged to the same parent and after its position
+            //  it will be removed from the children list
+            //  in that case we need to adjust the following elements target position
+            if (parent === target && index <= lastPos.index) {
+              // the next elements will be inserted 1 element before this one
+              domPositionOffset++;
+            }
+          });
+      }
     }
 
     if (this.plh) this.plh.style.display = 'none';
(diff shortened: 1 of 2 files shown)


lexoyo:
Here is what I did:

* when the user drags multiple elements, reorder the dragged elements to reflect the elements positions in the dom - needs to handle the case where the elements are not in the same container
* when the user drops multiple elements, move them to the right place, i.e. on top of the previously dropped one

I'd appreciate any feedback on the code or tests to make sure I didn't break anything @japo32 ?

Also I did not create any tests because I couldn't find a suitable place, should I @artf ?

Here is how it behaves now:

Uploading 2022-03-29 12-50-56.mp4…

artf (changes requested):
Hi @lexoyo thanks for the PR.
I'd say I'd not expect any change in `MoveComponent` and I'd not rely on DOM for sorting as it might differ from the component tree (the component view can be completely custom).

I think this issue might be solved on the sorter level by relying on the component methods like `component.index()` and `component.parent()`.

Here what I'd expect from a simple sort by component index.
```js
const toMoveArr = ...;

if (toMoveArr.length > 1) {
  toMoveArr.sort((a, b) => {
    // inverted sort as we append always to the same lastPos
    return b.index() - a.index();
  })
}

toMoveArr.forEach(model => { ...
```
This obviously doesn't count the sort across different parents which is probably the main reason I always postponed this fix 😅

lexoyo:
Hi @artf , thx for the quick reply

> Hi @lexoyo thanks for the PR. I'd say I'd not expect any change in `MoveComponent` and I'd not rely on DOM for sorting as it might differ from the component tree (the component view can be completely custom).

It makes sense of course !

> I think this issue might be solved on the sorter level by relying on the component methods like `component.index()` and `component.parent()`.

Ok, very feasable, I did this

> Here what I'd expect from a simple sort by component index.
> 
> ```js
> const toMoveArr = ...;
> 
> if (toMoveArr.length > 1) {
>   toMoveArr.sort((a, b) => {
>     // inverted sort as we append always to the same lastPos
>     return b.index() - a.index();
>   })
> }
> 
> toMoveArr.forEach(model => { ...
> ```

I did it like this, but then the problem is when we drop in the same container, after the elements, they get inverted

This is why this commit is needed
https://github.com/artf/grapesjs/pull/4220/commits/d0b37a0c3b1903644856d6683516671fb99b95a1

You can test it here, I am still testing it too
https://lexoyo.me/grapesjs/demo.html

(Sorry for the useless mess in some files, due to linter, I'll clean it at one point so that I don't change files I did not touch)

src/commands/view/MoveComponent.js
```diff
@@ -87,6 +87,8 @@ export default extend({}, SelectPosition, SelectComponent, {
    * @private
    */
   initSorterFromModels(models) {
+    // todo: take common ancester instead of parent
+    const _frame = (this.em.get('currentFrame') || {}).model;
```
artf, line 91: I guess this one is to remove

test/specs/utils/Sorter.js
```diff
@@ -522,14 +514,75 @@ describe('Sorter', () => {
...
+      expect(obj.sort(withParents(child00), withParents(child10))).toEqual(1);
+      expect(obj.sort(withParents(child1), withParents(child2))).toEqual(1);
+      expect(obj.sort(withParents(child10), withParents(child2))).toEqual(1);
+
+      expect(obj.sort(withParents(child1), withParents(child00))).toEqual(-1);
+      expect(obj.sort(withParents(child01), withParents(child00))).toEqual(-1);
+      expect(obj.sort(withParents(child10), withParents(child00))).toEqual(-1);
+      expect(obj.sort(withParents(child2), withParents(child1))).toEqual(-1);
+      expect(obj.sort(withParents(child2), withParents(child10))).toEqual(-1);
+    });
```
artf, line 586: ❤️

src/utils/Sorter.js
```diff
@@ -982,6 +982,35 @@ export default Backbone.View.extend({
...
+      throw new Error(
+        `Could not find a common ancester for the dragged elements #${obj1.model.getId()} and #${obj2.model.getId()}`
+      );
+    // find siblings in the common ancester
+    // the sibling is the element inside the ancester
+    const s1 = obj1.parents[obj1.parents.indexOf(ancester) - 1];
+    const s2 = obj2.parents[obj2.parents.indexOf(ancester) - 1];
+    // order according to the position in the DOM
+    return s2.index() - s1.index();
+  },
```
artf, line 1012: Nice ❤️
I'm only wondering if `if (!ancester)` is possible and if we can at least fallback to a simple `obj2.model.index() - obj1.model.index()` instead of throwing error

src/utils/Sorter.js
```diff
@@ -1013,9 +1042,36 @@ export default Backbone.View.extend({
     if (this.moved && target) {
       const toMove = this.toMove;
       const toMoveArr = isArray(toMove) ? toMove : toMove ? [toMove] : [src];
-      toMoveArr.forEach(model => {
-        moved.push(this.move(target, model, lastPos));
-      });
+      let domPositionOffset = 0;
+      toMoveArr
```
artf, line 1046: There are cases for the sorter where toMoveArr is `[undefined]`, which allows the drop from blocks, native D&D and sort of layers in Style Manager (you can see instead how these are not working properly). This is why I was suggesting to use something `toMoveArr.length > 1` or any other check to prevent sorting 😅

artf (commented):
Overall looks to work pretty well 👏

lexoyo:
@artf I took into account your review

I reverted `MoveComponent.js` file to its version before I changed it

For your last 2 comments, I think I got it, please confirm. I have tested dropping images, now it works :))

lexoyo:
also i updated https://lexoyo.me/grapesjs/

artf (approved):

artf:
Awesome, thank you @lexoyo ❤️

PR #4238, opened 2022-04-04
fix in docs

lexoyo:
Am I mistaken ?

### DIFF
--- a/docs/modules/Storage.md
+++ b/docs/modules/Storage.md
@@ -269,11 +269,11 @@ Storage.add('remote-local', {
     const remoteStorage = Storage.get('remote');
 
     try {
-      await remoteStorage.store(data, Storage.getStorageOptions('remote'));
+      await remoteStorage.store(data, Storage.getConfig().options.remote);
     } catch (err) {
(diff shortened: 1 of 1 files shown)


Ju99ernaut (commented):
In this context you'd need to pass the options as previously shown, these are no longer passed automatically as done in previous versions. Also note that these are different from `editor.store(data)` which still passes these options automatically.

artf:
Yeah, the example is correct as you're calling the built-in storage itself which doesn't have default options if not in the config object.

When the storage is called from the Storage Manager module, default config options are injected automatically.
```js
Storage.add('remote-local', {
  async store(data, options) {
    // `options` here will be equal to `Storage.getStorageOptions('remote-local')`
    // You can always enforce some default in your implementation
    const value = options.someProp || 'defaultValue';
    // ...
  },
  // ...
});
```

The consumer of the custom storage will be able to adjust default options in their configuration.
```js
grapesjs.init({
 // ...
 storageManager: {
   options: {
    'remote-local': { someProp: 'value'  }
   }
 }
})
```

PR #4241, opened 2022-04-05
TypeScript migration kickoff

xQwexx:
So I started to do the typescript migration. And I think the first we need to figure out is the base classes which will help to make this migration more smoother and to guide it to the end it will be a universal design.

In this pr contains:
- The updated base classes (I put them a different folder so we could migrate partially)
- Updated test env so it can run with partial ts environment

Next I plane to migrate the TraitsManager just first I would like to get some feedback on this design patterns :)


artf (commented):
Hi @xQwexx thanks for this PR, I really appreciate the effort.
I'd say overall LGTM, there are only a few points I'd like to have a better understanding of the proposal.
* I like the idea of switching Modules to classes (something I was actually already planning to do) but I'm curious about the Configuration object being a class. Is it a problem to keep the module configuration as a simple typed object?
* I know there are a few places where I need the Module in some Models, but is it correct to bind the generic Model to a Module? Maybe it would make sense to use some kind of ModuleModel where the module is required (or even simply required in the model properties)

xQwexx:
@artf Thank you for the fast response :)
- I moved around some staff so in the beginning it is made sense to make the config a class but yeah it is not anymore. So I updated it.
- As for the second question currently I think all or the most model get the {em, config} parameter pair in initialisation so instead that it will receive the module so it can use it for get the moduleConfig and to communicate with the editorModel  thought it. So this enable to add some intermediate function later, like a logger which also adds automatically the module name for the message. Or make it more module dependent so they don't call the EditorModel rather they call the intended module directly so we would be clear the modules dependencies but this probably be later or not done at all. I am not sure that it makes sense or should I explain it further? I also started to convert the Editor module into this structure to demonstrate.

artf (commented):
Ok, I'd say what you're saying makes sense to me (the logger is a good example).
I'd only ask you to remove `dist/grapes.min.js` and `dist/grapes.min.js.map` from the PR.

xQwexx:
Done :)

artf (approved):

PR #4259, opened 2022-04-13
Fix on generated css for transform property

codingaddicted:
### Issue description
When you apply a transform by using not all parameters, the generated css had a syntax like 
`transform: rotateX(1) 0 0 1 1 1`
which is not correct for browsers.
image

Forcing the function interpolation (when needed), the same value is generated as
image
which is a valid syntax.

### Changelog
Applied the transform function even on default value to get the working css syntax.
Added a new parameter on property to conditionally apply the function

### DIFF
--- a/src/style_manager/model/Property.js
+++ b/src/style_manager/model/Property.js
@@ -364,7 +364,8 @@ export default class Property extends Model {
    */
   getFullValue(val) {
     const fn = this.get('functionName');
-    const def = this.getDefaultValue();
+    const fnOnDef = this.get('applyFunctionOnDefault');
+    const def = fn && fnOnDef ? `${fn}(${this.getDefaultValue()})` : this.getDefaultValue();
     let value = isUndefined(val) ? this.get('value') : val;
     const hasValue = !isUndefined(value) && value !== '';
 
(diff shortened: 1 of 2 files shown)


artf (changes requested):
Thanks @codingaddicted for the PR but in all honesty I'd like to avoid adding a new property just to handle one exception.
Unfortunately, this is my fault for poorly implementing the `transform` property. It should be of type `stack` (similar to `transition` below) and not `composite`.

If you're willing to fix it by moving it to a `stack` type, I'll be happy to merge it. In case you need a more advanced usage of the stack properties usage I'd recommend checking out the [grapesjs-style-filter](https://github.com/artf/grapesjs-style-filter/blob/master/src/index.js) plugin code.

codingaddicted:
I totally agree with you that a property for a single use (actually there are two of them, but the "url" function is not affected by the syntax issue) can be an avoidable change, but it was introduced to get the component working without having side-effects (on background-image property, which is using the url function).

Since there's a type incompatibility on value (number) and its working default (string with function) it's not possible to make a choice between the two cases without an external (configuration) condition.

Anyway, I will look at the stack type option.

PR #4264, opened 2022-04-17
Convert EditorModule files into ts

xQwexx:


src/editor/config/config.ts
```diff
@@ -1,4 +1,5 @@
 export default {
+  name: 'Editor',
```
artf, line 2: Why do we need the `name` here?

src/asset_manager/index.js
```diff
@@ -321,7 +321,7 @@ export default () => {
      * ));
      */
     render(assts) {
-      if (this.getConfig('custom')) return;
+      if (this.getConfig().custom) return;
```
artf, line 324: I'm wondering if all these refactors in getConfig are purely a preference choice or there is something else? Aren't we able to support both options?

artf (commented):
Nice @xQwexx 👍
I have only a few questions

src/asset_manager/index.js
Vac1911: Not the contributor but it would be very difficult (maybe impossible) to provide dynamic type bindings for the function based on the input string. 

For example here is the current typescript binding for the BlockManager:
```js
interface BlockManager {
  getConfig(): BlockManagerConfig;
}

interface BlockManagerConfig {
    appendTo?: HTMLElement | string;
    blocks: Block[];
}
```
If I were to do `let foo = bm.getConfig().blocks` typescript would 100% know that `foo` is a `Block[]`. But there is no easy way (atleast that i know of) to change the expected return type based on the parameter passed.

Vac1911: Following up on this, there is actually a way to do this using [Function Overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads), but it is impractical.

For the example above you could theoretically do something like this:
```js
getConfig(): BlockManagerConfig
getConfig(key: 'appendTo'): HTMLElement | string | null
getConfig(key: 'blocks'): Block[]
getConfig(key?: 'blocks' | 'appendTo') {
  // code goes here
}
```

You would have to declare the return type for each key. While I don't recommend this approach, I wanted to mention I just found out that it is technically possible.

xQwexx: So I would get away from the getConfig() function, and use the config property instead, I am not sure how the best way to deprecate it, I think when we convert everything to typescript we should get ride of the deprecated functions as well. but I can live the  getConfig(key: string): any function but we get ride off the usages of it.

src/editor/index.ts
```diff
@@ -651,17 +735,7 @@ export default class EditorModule {
...
-        this[prop[0]] = 0;
-      } else {
-        this[prop] = 0;
-      }
-    });
-    this.modules = 0;
-    this.editorView = 0;
-    this.em = 0;
-    this.c = 0;
+    this.editorView = undefined;
```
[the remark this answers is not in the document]
xQwexx: This is not needed because the garbage collector will solve this, also in ts it isn't possible to assign undefined variables which types are not ?: .

src/asset_manager/index.js
artf: @xQwexx sounds good to me but yeah for now I'd prefer to keep it to reduce any possible breaking change.

@Vac1911 I was thinking of something like
```ts
getConfig<
  P extends keyof BlockManagerConfig | undefined = undefined, 
  R = P extends keyof BlockManagerConfig ? BlockManagerConfig[P] : BlockManagerConfig
>(prop?: P): R {
  const { config } = this;
  return prop ? config[prop] : config;
}
```

PR #4267, opened 2022-04-18
Convert transform to stack property

codingaddicted:
Transform property is now a stack property.
Added support for translation of elements.

Value step is defaulted as following:
- scaleX/Y/Z: step 0.01
- rotateX/Y/Z: step 1
- translateX/Y: step 1


artf (approved):
Nice thanks @codingaddicted

PR #4298, opened 2022-04-29
Update Assets.md

richardmtheobald:


artf (changes requested):
Thanks @richardmtheobald but this is not correct. The custom asset manager shouldn't handle the selection, this is part of the caller to decide how the selection logic works and this is why on the custom event you have `props.select` callback. 
Here is an example of how the asset manager works on the Image component.
```js
am.open({
  select(asset, complete) {
    imageComponent.set({ src: asset.getSrc() });
    complete && am.close();
  },
});
```
So, if you're implementing a custom Asset Manager, it's not up to you to decide how to apply the selected asset, your part should only be to pass the selected asset to the select prop: `props.select(selectedAsset)`

PR #4306, opened 2022-05-04
Convert Selector Module into ts

xQwexx:


artf (approved):
Awesome ❤️

FYI I'm working on the LayerManager refactoring so leave it to me for the TS migration

PR #4309, opened 2022-05-05
Added eslint support for ts file and enable some eslint rule

Singwai:
Context:
I added eslint parser to support  linting for ts file. I also enabled eslint rule `qoutes` and `eol-last` as warning. They were previously disabled. If we are happy with these changes, I will apply the autofix in a separate PR.

This should also remove the need for helper script `format` in package.json in the future

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -9,9 +9,13 @@
   },
   "rules": {
     "strict": 0,
-    "quotes": [0, "single"],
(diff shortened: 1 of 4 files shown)


artf (approved):

artf:
Nice 👍  thank you @Singwai

artf:
@Singwai fixes are pushed

PR #4312, opened 2022-05-07
Start of the Components ts refactor

xQwexx:


artf (changes requested):
All good, I'd only ask you to place Backbone related properties (`tagName`, `events`) as methods or at the prototype level as its extend mechanism conflicts with ES6 classes properties (eg. you can see how it's not possible to activate image and text components by a double click).
```js
class View extends BaseView {
  tagName() {
   return '...';
  }
  events() {
    return {...}
  }
}
// or
View.prototype.tagName = '...';
View.prototype.events = {...};
```

src/dom_components/index.ts
```diff
@@ -417,23 +426,35 @@ export default class ComponentManager extends Module {
...
 
     // Function for extending source object methods
-    const getExtendedObj = (fns, target, srcToExt) =>
+    const getExtendedObj = (fns: any[], target: any, srcToExt: any) =>
       fns.reduce((res, next) => {
         const fn = target[next];
         const parentFn = srcToExt.prototype[next];
         if (fn && parentFn) {
-          res[next] = function (...args) {
+          res[next] = (...args: any[]) => {
```
artf, line 457: Be careful with arrow functions as `this` inside the function changes its meaning (I'll fix it)

artf (approved):

PR #4313, opened 2022-05-09
updated mousedown to click event for consistency

m-jojo-s:


artf (commented):
Unfortunately, they are issues with the drag/move toolbar button and the click event, so at the moment this workaround is necessary.

PR #4314, opened 2022-05-09
Link to English Wikipedia

edemaine:
Given that the README is in English, it makes sense to link to English Wikipedia instead of Italian. At least I found it surprising to be brought to Italian Wikipedia. 🙂


artf (approved):
Ops 😁, thanks

PR #4316, opened 2022-05-10
Convert the rest of the Canvas module into ts

xQwexx:


src/canvas/view/FrameView.ts
```diff
@@ -305,18 +321,18 @@ export default class FrameView extends View {
...
   renderBody() {
-    const { config, model, ppfx } = this;
-    const { em } = config;
+    const { config, em, model, ppfx } = this;
     const doc = this.getDoc();
     const body = this.getBody();
     const win = this.getWindow();
-    const conf = em.get('Config');
+    const conf = em.config;
+    //@ts-ignore TODO I don't understand why this needed nowhere else is used
```
artf, line 335: The idea was to make use of it inside custom components with scripts so you could change the behavior of components during the editing (eg. disable JS animations if the element is in the editor) but in all honesty, I'm not sure anymore if it's the right approach. For now, let's leave it here.

xQwexx:
I am not sure what is the problem if I run with npm run build:js it looks completely fine, and I did not get any errors in the browser. So I am not sure how can I fix the issue if I am unable to reproduce it. :(

artf:
This is what I see if I revert the revert.
Schermata 2022-05-13 alle 13 14 43

From a quick check, I see how some canvas classes are wrong and there is also missing the hover tool (so I'm not able to see the component outline border on hover).
Schermata 2022-05-13 alle 13 17 31
Correct structure
Schermata 2022-05-13 alle 13 19 28

Aren't you able to reproduce it (maybe on a temp branch)?
**Worth mentioning, if you pull from the dev now to the canvas-module branch, your work that was reverted will be gone. You should revert the revert commit (git revert 6ddb5f8e9e8f4d5c09c6aae5738a5c5ef4f3acbf).**

xQwexx:
I am getting the correct structure with just reverting the changes:
Screenshot 2022-05-13 at 14 49 47
I am not sure what could be the problem
Are you using the default editor from this repo or more custom one,
I thinking that it could have some different in the config parameters

artf:
Ok, I think I got it why you can't reproduce it.
I forgot to update the grapesjs JS file in `index.html` 😅
Right now it's loading the static `<script src="dist/grapes.min.js"></script>`, this should be replaced with the JS file from the dev server `<script src="grapes.min.js"></script>`

PR #4324, opened 2022-05-12
Convert Traits module into ts

xQwexx:
I am planning to do a bigger refactor overhaul of the traits just first I want to have as typescript


index.d.ts
```diff
@@ -516,7 +516,7 @@ declare namespace grapesjs {
 
   interface TraitView {
     noLabel?: boolean;
-    templateInput?: string | ((options: { trait: Trait }) => string);
+    templateInput?: string;
```
artf, line 519: I see what you have done in View classes but we can't remove the support of `templateInput` as a function as it's necessary for the current customization mechanism and it is also publicly documented.

artf, line 519: We can't remove the support of `templateInput` as a function as it's necessary for the current customization mechanism.

src/trait_manager/view/TraitView.ts
```diff
@@ -92,42 +112,27 @@ export default class TraitView extends Backbone.View {
...
-      tpl =
-        this.createLabel({
-          label,
-          component: target,
-          trait: this,
-        }) || '';
-    }
+  private renderLabel() {
+    const { $el } = this;
+    let tpl = this.templateLabel;
```
artf, line 135: I know it doesn't make sense from the class perspective to have undefined methods but we can't simply remove methods publicly documented for customization like `createLabel`
https://grapesjs.com/docs/modules/Traits.html#update-layout

artf, line 135: I know it doesn't make sense from the class perspective to have undefined methods but we can't simply remove methods publicly documented for customization like `createLabel`
https://grapesjs.com/docs/modules/Traits.html#update-layout

artf (changes requested):

index.d.ts
xQwexx: I added it back

src/trait_manager/view/TraitView.ts
xQwexx: I reverted it but I think these are redundant capabilities so I would like to get ride of them sometimes

src/trait_manager/index.ts
```diff
@@ -0,0 +1,151 @@
...
+        config,
+      },
+      this.getTypes()
+    );
+    this.view = view;
+    return view.el;
+  }
+
+  destroy() {
+    //this.__destroy();
```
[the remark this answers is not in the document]
xQwexx: Yes it was intentional I also deleted the destroy function

artf:
Thanks @xQwexx can you please resolve pending conflicts so I'll be able to review it again?

artf:
I'm closing this one as traits module is now converted to ts

PR #4329, opened 2022-05-15
Convert Storage module to ts

xQwexx:


src/abstract/Module.ts
```diff
@@ -1,7 +1,7 @@
-import { isElement, isUndefined } from 'underscore';
-import { Collection, View } from '../common';
-import EditorModel from '../editor/model/Editor';
-import { createId, isDef } from '../utils/mixins';
+import { isElement, isUndefined } from "underscore";
```
artf, line 1: Aren't you able to fix the issue with quotes so we can suppress all the warnings in the PR?
In case you can't I'd ask you to run `npm run lint -- --fix`

artf (changes requested):

src/abstract/Module.ts
Singwai: I am including `ts` files on lint-staged to auto-fix this kind of lining issue before committing in the future. 
https://github.com/artf/grapesjs/pull/4375

src/utils/fetch.ts
```diff
@@ -1,23 +1,31 @@
...
-export default typeof fetch == 'function'
-  ? fetch.bind()
-  : (url, options) => {
-      return new Promise((res, rej) => {
+interface FetchOptions {
+  method: "POST" | "GET";
+  credentials: string;
+  headers?: { [key: string]: string };
+  onProgress: ((this: XMLHttpRequest, ev: ProgressEvent) => any) | null;
+  body?: Document | BodyInit | null;
```
Vac1911, line 13: I believe this should be changed to `Document | XMLHttpRequestBodyInit | null`. 

This option is being passed to `XMLHttpRequest.send(body?: Document | XMLHttpRequestBodyInit | null)` on line 42. `BodyInit` is a union type of `ReadableStream` and `XMLHttpRequestBodyInit`, so `BodyInit` is not assignable to `XMLHttpRequestBodyInit`.

artf:
I'm closing this one as storage module is now converted to ts

Issue #3770, opened 2021-09-12
BUG: Multi-Selection Drag issue with dragMode: absolute


damnslow:
### GrapesJS version

- [X] I confirm to use the latest version of GrapesJS

### What browser are you using?

Chrome v93

### Reproducible demo link

https://codepen.io/damnslow/pen/yLXXOxQ

### Describe the bug

**How to reproduce the bug?**
1. Select both the elements
2. Drag using drag icon

**What is the expected behavior?**
Both elements should move together

**What is the current behavior?**
Only one element moves.

If is necessary to execute some code in order to reproduce the bug, paste it here below:
No.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

damnslow:
Hi @artf, is there any workaround to solve this?

artf:
Absolute positioning, right now, has more than one issue and I'd like to address them all here:
- [ ] Wrong position on a different Devices
- [ ] Wrong position on a zoomed/panned canvas
- [ ] Support touch devices

I don't know exactly when I'll start working on it but I hope soon, and as always, any help is appreciated.

ahmet1340:
Did you find time to deal with it? I'm currently working on a drag and drop project and I'm having an issue with mispositioning on different devices also I can't use the setzoom function for mobile. I would be very grateful if you would take care of this issue. (Google translate used.)

Adham380:
These issues have still not been resolved and are frankly annoying. Changing the absolute dragging to a % instead of a px offset makes so much more sense and solves most issues, but for some reason wasn't chosen. Is there any plans to change this?

artf:
@Adham380 PRs are welcome

dalmatele:
Agree. It also affects the menu of NavBar in Mobile mode. If you set dragMode to "absolute" and move items in NavBar then switch to Mobile Preview mode, you'll see NavBar does not work.

vizardkill:
up

voleep:
Up

fmfeit:
Hi @artf,
this feature is crucial for my current project, so I have two questions:
1. Is it possible to pay you (or someone else) to prioritize this feature?
2. If not, can you outline the steps necessary to resolve the issue and give me a starting point?

artf:
@fmfeit here you can find the related source of the [command](https://github.com/GrapesJS/grapesjs/blob/dev/src/commands/view/ComponentDrag.ts).

itscarlosrufo:
[...]

PR #4343, opened 2022-05-25
Use coordinates of drag onEnd event for placement in absolute mode. 

contentfree:
Uses clientX/Y from the dragger's onEnd event. Relates to #3770. While it doesn't handle zoom, it should correctly handle a panned canvas.

### DIFF
--- a/src/utils/Droppable.js
+++ b/src/utils/Droppable.js
@@ -112,8 +112,8 @@ export default class Droppable {
           let comp;
           if (!cancelled) {
             comp = wrapper.append(content)[0];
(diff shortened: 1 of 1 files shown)


contentfree:
Looks like some other things have changed in the `dev` branch that lead to the initial drop event not firing in the same way as when I was testing from the unpkg version. I'll update this PR when I figure that out.

contentfree:
The drag end event seems to be using the _first_ drag event from entering the canvas… I'm going to close this pull request as this isn't working with the current dev. It *did* work with 0.18.4… not sure what has changed since then.

PR #4363, opened 2022-06-02
add option readonly for InputTrait

vinicius-actcon:
NumberTrait has built in custom "stepper" mecanism cobined with configurations such as [min, max, step] that works pretty well when used isolated, but the user can insert manually values into the input that not respect this configurations, so with this flag we can force to the user use only the "arrows" controls instead of edit the input manually.

Thanks!


artf (changes requested):
I guess it's worth exploring why/where the input doesn't respect those options (min/max/etc) as this kind of option looks quite confusing. Honestly, with a `readonly` option I wouldn't expect either the arrows to change the value.

artf:
Closing for inactivity

PR #4373, opened 2022-06-10
Fix handle default svg properly in component image when reload from local storage

Singwai:
Added ts support for jest
Removed relative src import in test directory
Fix `isDefaultSrc` and `src` initialization logic

Context:
When we export an image component with a default SVG image and re-import it back to grapesjs (using JSON or HTML), grapesjs won't recognize the image URL as the default src because when grpaesjs exports svg image, it converts the svg to a base 64 string.

This PR extracts the b64 logic to a util function and checks the b64 version of the default src when we initialize an image component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "@babel/cli": "^7.15.7",
     "@babel/preset-typescript": "^7.16.7",
     "@types/backbone": "^1.4.15",
+    "@types/jest": "^28.1.1",
     "@typescript-eslint/parser": "^5.22.0",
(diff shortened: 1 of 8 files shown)


artf (approved):

artf:
Awesome @Singwai thanks also for providing tests ❤️

skru:
Since this change, the gjs-plh-image class is added to any image block we have when dragging into the canvas/loading from local storage (indexdeb plugin).

Is that intentional?

The styling for .gjs-plh-image is causing issues for us

All the best!

Issue #2824, opened 2020-06-11
Dragging items do not position on intital drag?


tonypartridge:
So when I drag a component/item from the bar to the center it will not go where I told it to go... IT always goes about 20% page width to one side? Once added it's adjustable. See this gif:
https://gyazo.com/87076de55a1853e80b5dd7ea14ccb94e

Does anyone have any suggestions?

RJCAM:
Hi @tonypartridge 
Do you have the last GrapesJS version?
This is working fine for me with the `dragMode: "absolute",`
Check this basic code below:

```
<!DOCTYPE html>
<html lang="en">

<head>
    <meta charset="utf-8">
    <title>GrapesJS</title>
    <link href="https://unpkg.com/grapesjs/dist/css/grapes.min.css" rel="stylesheet">
    <script src="https://unpkg.com/grapesjs"></script>
    <style>
        body,
        html {
            height: 100%;
            margin: 0;
        }
    </style>
</head>

<body>

    <div id="gjs" style="height:0px; overflow:hidden"></div>

    <script type="text/javascript">
        var editor = grapesjs.init({
            height: '100%',
            dragMode: "absolute",
            noticeOnUnload: 0,
            storageManager: {
                autoload: 0
            },
            container: '#gjs',
            fromElement: true,
            plugins: [],
            pluginsOpts: {}
        });

        editor.BlockManager.add('test-block', {
            label: 'Div',
            attributes: {
                class: 'gjs-fonts gjs-f-b1',
            },
            content: `<div class="test">This is a simple div</div>
            <style>
                .test {
                    width: 200px;
                    height: 30px;
                    text-align: center;
                    background-color: lightgrey;
                }
            </style>`,
        });
    </script>
</body>

</html>
```

artf:
yeah... I can actually reproduce the same issue when a smaller device view is enabled, probably some kind of regression. I need time to investigate but at the moment I'm kind of busy with other stuff and dragMode is not my priority right now. So, I hope someone else could look at this

soltanloo:
I have the same issue when I'm using the absolute drag mode. Blocks always fall out of the canvas and I need to scroll to grab them and place them back into the original desired position.

PR #4387, opened 2022-06-15
Take canvas offsets into account when dropping a new block on the canvas. 

contentfree:
Relates to #2824 / #3770

I'm not sure this is how you'll want to solve this globally, @artf, but it does seem to fix the dropping of blocks in designer mode on a narrow canvas. I believe there's similar issues with the initial position of the dragged element: Canvas.getMouseRelativeCanvas doesn't correctly take into account the canvas offset but I think the main cause may be CanvasView.getPosition and the way it combines frame offset & canvas offset in a strange way.

### DIFF
--- a/src/utils/Droppable.js
+++ b/src/utils/Droppable.js
@@ -112,8 +112,13 @@ export default class Droppable {
           let comp;
           if (!cancelled) {
             comp = wrapper.append(content)[0];
-            const { left, top, position } = target.getStyle();
-            comp.addStyle({ left, top, position });
+            const canvasOffset = editor.Canvas.getOffset();
+            const { top, left, position } = target.getStyle();
+            comp.addStyle({
+              left: parseFloat(left) - canvasOffset.left + 'px',
+              top: parseFloat(top) - canvasOffset.top + 'px',
+              position,
+            });
           }
           this.handleDragEnd(comp, dt);
           target.remove();


contentfree:
Also, I'm a grapesjs newb and it's likely there's a more elegant way to solve this more broadly.

artf (approved):
I guess there is a lot of stuff to check/improve for the absolute mode but at least I see how it fixes the current issue, so at the moment is better than nothing.

Thank you 🙇

contentfree:
Yeah, it's a bit brute force for this specific fix. I'm very new to GrapesJS but willing to take another crack at it if you've got some guidance on where a better and broader fix would be applied.

artf:
Honestly, I'm not sure if this is something I'd suggest to someone new to GrapesJS 😅 
There are different places where the D&D of components can start (from blocks, inside the iframe, native HTML5 D&D outside the editor) and in each case, there are different constraints and you have to make it work in different modes (absolute/translate/standard).

contentfree:
Strangely, this line `const canvasOffset = editor.Canvas.getOffset();` works for me locally but doesn't once incorporated to grapesjs. I might need to push an update to you that does something like `em.get("Canvas")` instead.

artf:
@contentfree you're right, my bad I didn't notice it. I'll push a fix

PR #4427, opened 2022-07-07
Fix broken link for Editor configuration Object

Ashwinvalento:
Editor configuration Object was broken after migrating to typescript


artf:
Thanks @Ashwinvalento the actual fix should be made here https://github.com/artf/grapesjs/blob/22fb44ba9934f87ce12f60fa2a7b37b6ec37835b/src/editor/index.ts#L3
as the API .md files are generated from the source.

artf:
Fixed the source file here https://github.com/artf/grapesjs/commit/fa64dce631a9f71a84abd5fb177851a5361c3a4a

PR #4431, opened 2022-07-08
Convert Panel module into ts

xQwexx:
I had some time so I converted this module into ts :)
Also I made the Collection interface nicer

I updated the backbone dependency to 1.4.1 because 1.3.3 didn't contained the preinitialise() and I needed that to make this generic Collection.

### DIFF
--- a/src/abstract/Collection.ts
+++ b/src/abstract/Collection.ts
@@ -1,5 +1,5 @@
 import Backbone, { AddOptions } from 'backbone';
-import { isArray, isObject } from 'underscore';
+import { isArray, isObject, isUndefined } from 'underscore';
 import Model from './Model';
 
 type Module<TModel extends Model> = TModel extends Model<infer M> ? M : unknown;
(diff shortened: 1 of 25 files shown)


src/abstract/Collection.ts
```diff
@@ -1,6 +1,35 @@
...
+  module!: Module<TModel>;
+  private newModel!: ModelConstructor<TModel>;
+
+  add(model: Array<Record<string, any>> | TModel, options?: AddOptions): TModel;
+  add(models: Array<Array<Record<string, any>> | TModel>, options?: AddOptions): TModel[];
+  add(model?: unknown, options?: AddOptions): any {
+    //Note: the undefined case needed because backbonejs not handle the reset() correctly
+    var models = isArray(model) ? model : !isUndefined(model) ? [model] : undefined;
+
+    models = models?.map(m => (m instanceof this.newModel ? m : new this.newModel(this.module, m))) ?? [undefined];
```
artf, line 18: IDK why but when I try to load the editor with these changes it seems to break at this line
Schermata 2022-07-15 alle 13 49 12
Looks like `this.newModel` is undefined

artf (changes requested):

src/abstract/Collection.ts
xQwexx: I think you forgot to update the dependency. Sorry forgot to write it in the description (Updated)

artf: Yeah, you're right, after update, it seems to work properly 👍

artf (approved):

PR #4433, opened 2022-07-09
Convert Modal module into ts

xQwexx:


src/modal_dialog/index.ts
```diff
@@ -0,0 +1,264 @@
...
+  getTitle() {
+    return this.model.get('title');
+  }
+
+  /**
+   * Set the content of the modal window
+   * @param {string | HTMLElement} content Content
+   * @returns {this}
+   * @example
+   * // pass a string
```
Singwai, line 202: This example is kind of useless after migrating to ts. 

Consider removing it.

src/modal_dialog/index.ts
```diff
@@ -0,0 +1,264 @@
...
+   * @param {Object} config Configurations
+   * @private
+   */
+  constructor(em: EditorModel) {
+    super(em, 'Modal', defaults);
+
+    this.model = new ModalM(this);
+    this.model.on('change:open', (enable: boolean) => {
+      em.trigger(`modal:${enable ? 'open' : 'close'}`);
+    });
```
artf, line 59: The first argument of the callback is the model itself and the value is the second one.

artf (approved):
Nice, thanks @xQwexx 
I found a tiny issue with the open modal event but I'll take care of it.

PR #4439, opened 2022-07-11
Make Component valid child for Backbone.Model

xQwexx:
Because the Component is not a valid child for the Model we could not define views for it in typescript.

To fix it we need to get ride of the _validate() function which is a private function so it should not be overwritten.
This means loss of onBeforeStyle functionality which is a global css override I also think this functionality is unneeded, because its result more or less could be achieved with the usage of the global CSS file. (Which would make the changes easier to manage)


xQwexx:
I also need to something with the defaults, I didn't figured that out yet

artf (changes requested):
Thanks @xQwexx overall looks good but currently seems like there is an issue when you try to add a new custom component due to the new `defaults` getter, eg.
```js
editor.DomComponents.addType('all-traits', {
  model: {
    // defaults might also be a function
    defaults: {...}
  },
});
```
Schermata 2022-07-19 alle 15 57 04

xQwexx:
I tested it, so it should work now

artf:
Thanks @xQwexx I really appreciate these changes ❤️

PR #4487, opened 2022-07-30
Use em.get instead of referencing editor

contentfree:
Not sure how this one snuck by in my last PR but I shouldn't've referenced `editor.Canvas` but instead should've used `em.get('Canvas')`


artf:
Thanks @contentfree I updated it [here](https://github.com/artf/grapesjs/commit/9d1466157737e3e55cb9d65a4d7dc49d09c1f0ab) by reusing the canvas instance which was already available

PR #4505, opened 2022-08-10
Fix wrong describe type on EditorConfig dragMode

tyutony:
config dragMode describe would be 'translate' | 'absolute' and not boolean.
it cause wrong alert when use typescript and dragMode propriety.


artf (approved):
Thanks @tyutony

PR #4521, opened 2022-08-19
Add correct mobile first sort on CSS rules view

iagormoraes:
This PR fixes the canvas issue related with mobile first configuration, where only the CSS generated has the correct output, now with this fix we also can see the breakpoints being added in correct order.

Reported issues on:

https://github.com/artf/grapesjs/issues/2774
https://github.com/artf/grapesjs/issues/1996


artf (approved):

artf:
Awesome, thanks @iagormoraes

PR #4557, opened 2022-09-06
Fix link editor config

IRediTOTO:


artf:
Thanks @IRediTOTO I updated the source file here https://github.com/artf/grapesjs/commit/fa64dce631a9f71a84abd5fb177851a5361c3a4a as this file is autogenerated

PR #4678, opened 2022-10-21
Some typescript updates

froodian:
When I upgraded our org's grapesjs from 0.18.3 to 0.20.1, I had to make these typescript additions for our integration to continue functioning. I verified in the underlying implementation that I believe these all make sense.


artf:
Thanks Ian 🙏

Issue #4743, opened 2022-11-22
BUG: stylePrefix property not overiding the gjs in most elements with the gjs-class


JonathanRiche:
### GrapesJS version

- [X] I confirm to use the latest version of GrapesJS

### What browser are you using?

Version 107.0.5304.107 (Official Build) (64-bit)

### Reproducible demo link

https://jsfiddle.net/jonrobert/2gLdur9c/3/

### Describe the bug

**How to reproduce the bug?**
1.  Modify the stylePrefix property on any of the options in the config option when running the grapesjs.init() function
ie  
```js
layerManager:{
  	stylePrefix:'customStylePrefix'
  },
  blockManager: {
  stylePrefix:'customStylePrefix'        
 }
```

**What is the expected behavior?**
Expected behaviour is for the .gjs-classes are replaced with the stylePrefix like previous versions of grapesJs

**What is the current behavior?**
Current behaviour is the stylePrefix property is not being applied even when setting in the config

If is necessary to execute some code in order to reproduce the bug, paste it here below:
```js
const editor = grapesjs.init({
	//stylePrefix:'customStylePrefix',  
	container: '#gjs',
  fromElement: 1,
  height: '100%',
  storageManager: { type: 0 },
  plugins: ['gjs-blocks-basic'],
  layerManager:{
  	stylePrefix:'customStylePrefix'
  },
  blockManager: {
  stylePrefix:'customStylePrefix'        
 }
});
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #4753, opened 2022-11-29
Fix for issue #4743 added a check from config

JonathanRiche:
This fixes BUG: stylePrefix property not overiding the gjs in most elements with the gjs-class #4743


JonathanRiche:
This fixes/closes issue : BUG: stylePrefix property not overiding the gjs in most elements with the gjs-class #4743

artf (approved):
Thanks @JonathanRiche

PR #4849, opened 2023-01-16
Add missing colorPicker option properties in index.d.ts

KaoDeo:
ColorPicker options are described in editor config.ts but not in index.d.ts


artf (approved):

artf:
Thanks @KaoDeo

PR #4899, opened 2023-02-03
Convert Trait test into TS

xQwexx:
The ts config change needed to not use ../.../ in the test import


artf (changes requested):
Hi @xQwexx thanks for the PR, I'd suggest instead rolling back changes done in `tsconfig.json` and use `../../` as I noticed there might be some weird resolving issues when you try to alias paths.
I'd also ask you to rollback changes in the lock file as it seems to break tests.

artf (approved):

artf:
Thanks @xQwexx

PR #4903, opened 2023-02-05
fix for when no implicit type

wilsonsergio2500:
error when importing into project that use type safety.
'each' which lacks return-type annotation.

This tiny pr solves this challenge


artf (approved):

artf:
Thanks @wilsonsergio2500

PR #4910, opened 2023-02-06
Fix typing of trigger to include `args`

Zaxcoding:
Hello, I noticed that the typing for `editor.trigger` did not match between [`editor/index.ts`](https://github.com/GrapesJS/grapesjs/blob/a83847b1b35e69e5a5810ccf1eb1e1384bf0aab1/src/editor/index.ts#L761-L769) and the root `index.d.ts`, so I updated the `index.d.ts` to include the `args` param.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1243,7 +1243,7 @@ declare namespace grapesjs {
      * @param event - Event to trigger
      * @param args - payload to send along with the event
      */
-    trigger(event: EditorEvent, ...args: any[]): Editor;
+    trigger(event: GrapesEvent, ...args: any[]): Editor;
     /**
      * Destroy the editor
      */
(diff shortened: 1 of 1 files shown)


artf (changes requested):
Thanks @Zaxcoding this file is kind of temporary declaration and will be deprecated once we'll release the declaration generated directly from code, but at the moment I accept PRs here anyway.

In your changes you have to rollback `GrapesEvent` as there is no `EditorEvent` in this file.

Zaxcoding:
Sorry about that, I think `EditorEvent` must have come from the other definition for `trigger`.

artf (approved):

artf:
Thank you Zach

PR #4964, opened 2023-03-07
fix links in docs

lexoyo:


artf:
I was actually using incorrectly the anchor alias on the second link 😅 fixed here: https://github.com/GrapesJS/grapesjs/commit/eff8e20a49753895bab0116bda99081ac7381bb9
thanks for the callout though

PR #4963, opened 2023-03-07
Fix broken links in the docs

himansh-gjr:
Fixes #4962 

This is what I did

Changed javascript file links in configuration object to typescript files 
https://github.com/GrapesJS/grapesjs/blob/master/src/storage_manager/config/config.js => https://github.com/GrapesJS/grapesjs/blob/master/src/storage_manager/config/config.ts

https://github.com/GrapesJS/grapesjs/blob/master/src/block_manager/config/config.js => https://github.com/GrapesJS/grapesjs/blob/master/src/block_manager/config/config.ts

https://github.com/GrapesJS/grapesjs/blob/master/src/style_manager/config/config.js => https://github.com/GrapesJS/grapesjs/blob/master/src/style_manager/config/config.ts

https://github.com/GrapesJS/grapesjs/blob/dev/src/commands/config/config.js => https://github.com/GrapesJS/grapesjs/blob/master/src/commands/config/config.ts

https://github.com/GrapesJS/grapesjs/blob/master/src/asset_manager/config/config.js => https://github.com/GrapesJS/grapesjs/blob/master/src/asset_manager/config/config.ts

PR #4971, opened 2023-03-10
Update Style-manager.md config url

Eduardo1-10:
This url is broken https://github.com/GrapesJS/grapesjs/blob/master/src/style_manager/config/config.js. I'm thinking this is the right one: https://github.com/GrapesJS/grapesjs/blob/master/src/style_manager/config/config.ts


artf:
Merged already the [PR](https://github.com/GrapesJS/grapesjs/pull/4963) containing the same fix, but thanks for the callout

Issue #4972, opened 2023-03-12
BUG:  Checkbox trait not getting checked or unchecked on undo/redo actions.


FaisalShaikhHA:
### GrapesJS version

- [X] I confirm to use the latest version of GrapesJS

### What browser are you using?

Version 110.0.5481.177 (Official Build) (arm64)

### Reproducible demo link

https://grapesjs.com/demo.html

### Describe the bug

**What is the current behavior?**
Checkbox trait not getting checked or unchecked on undo/redo (trait value is updated on my end).
Reproducible on Grapesjs webpage [demo](https://grapesjs.com/demo.html).

**What is the expected behavior?**
The checkbox trait should get checked or unchecked on undo/redo.

https://user-images.githubusercontent.com/102785785/224533486-045ae63b-4134-4d3d-ac5c-8120404f0a1a.mp4

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #4983, opened 2023-03-17
Refactor Traits Collection

xQwexx:
I want to work on making the traits better, but first I wanted to tidy the current code.

### DIFF
--- a/src/trait_manager/index.ts
+++ b/src/trait_manager/index.ts
@@ -16,7 +16,7 @@ export const evAll = 'trait';
 export const evPfx = `${evAll}:`;
 export const evCustom = `${evPfx}custom`;
 
-const typesDef = {
+const typesDef: { [id: string]: { new (o: any): TraitView } } = {
   text: TraitView,
   number: TraitNumberView,
   select: TraitSelectView,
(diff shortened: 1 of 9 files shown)


src/trait_manager/index.ts
```diff
@@ -62,12 +62,9 @@ export default class TraitManager extends Module<TraitManagerConfig & { pStylePr
    */
   constructor(em: EditorModel) {
     super(em, 'TraitManager', defaults);
-    const c = this.config;
     const model = new Model();
     this.model = model;
-    const ppfx = c.pStylePrefix;
-    this.types = { ...typesDef };
-    ppfx && (c.stylePrefix = `${ppfx}${c.stylePrefix}`);
+    this.types = typesDef;
```
artf, line 67: Looks all legit, the only issue I see is with the removal of the that thing for `stylePrefix` which makes them render incorrectly. 
traits

Don't you see them in the same way?

xQwexx: Yeah thanks for catching it I didn't realised that it changed, I updated the fix for it.

src/trait_manager/model/Traits.ts
```diff
@@ -33,30 +37,25 @@ export default class Traits extends Collection<Trait> {
...
+        traits[i] = trait instanceof Trait ? trait : this.tf.build(trait);
+        traits[i].target = this.target;
       }
+      return super.add(traits, opt);
     }
+    const trait = models instanceof Trait ? models : this.tf.build(models as any);
+    trait.target = this.target;
 
-    return Collection.prototype.add.apply(this, [models as Trait[], opt]);
+    return super.add(trait, opt);
```
artf, line 59: I'm not sure exactly what is happening here but with these changes to Traits, undo/redo on traits doesn't work properly (eg. I tried my latest fix [here](https://github.com/GrapesJS/grapesjs/issues/4972) and it works only if I rollback these changes)

artf (changes requested):

src/trait_manager/model/Traits.ts
xQwexx: I added a test to cover the undo scenario and fixed it :) Sorry I couldn't have time earlier  :(

artf (approved):

artf:
Awesome, thanks @xQwexx ❤️

PR #4985, opened 2023-03-17
Module cleanup

xQwexx:
Small cleanup
Removing the init() function from the module, because it is not used anymore.
Fixing the warning with TS compiler:
image

### DIFF
--- a/src/editor/model/Editor.ts
+++ b/src/editor/model/Editor.ts
@@ -961,7 +961,7 @@ export default class EditorModel extends Model {
     // @ts-ignore
     const { editors = [] } = config.grapesjs || {};
     const shallow = this.get('shallow');
-    this._storageTimeout ?? clearTimeout(this._storageTimeout);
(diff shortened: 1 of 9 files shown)


src/editor/model/Editor.ts
```diff
@@ -975,7 +961,7 @@ export default class EditorModel extends Model {
     // @ts-ignore
     const { editors = [] } = config.grapesjs || {};
     const shallow = this.get('shallow');
-    clearTimeout(this._storageTimeout);
+    this._storageTimeout ?? clearTimeout(this._storageTimeout);
```
artf, line 964: I don't know if those TS issue you see (not sure why I'm don't see them on my side) are related to this part but with this change, by using `??` it looks like it won't be executed it case `_storageTimeout` is defined, did you mean to use `&&`?

xQwexx: And yes I wanted to use &&.

artf (approved):

PR #4990, opened 2023-03-20
Options while exporting code/template

AliHafeez337:
I had to provide "opts"
`editor.runCommand('core:open-code', opts)`
while exporting code but I guess its missing in "ExportTemplate.js" file...

The code on line 204
image
will run and pass options (args) 
which in turn runs this code:
image
but "getHtml" was not passing options...

### DIFF
--- a/src/commands/view/ExportTemplate.ts
+++ b/src/commands/view/ExportTemplate.ts
@@ -2,7 +2,7 @@ import { CommandObject } from './CommandAbstract';
 import { $ } from '../../common';
 
 export default {
-  run(editor, sender) {
+  run(editor, sender, opts = {}) {
     sender && sender.set && sender.set('active', 0);
     const config = editor.getConfig();
     const modal = editor.Modal;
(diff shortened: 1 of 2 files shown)


AliHafeez337 (commented):
I wanted to have no Id on code written inside "custom-code"... I passed "noId" to give a  general idea but I think a callback (returning true or false based on a component type) here would be more appropriate...

Thanks

artf:
Hey @AliHafeez337 [I improved a bit](https://github.com/GrapesJS/grapesjs/commit/9628fcf93beb1df205cd1aafc91d79768970f49f) your suggestion on adding options for the export template command, which I think might be useful also to others.

> I wanted to have no Id on code written inside "custom-code"... I passed "noId" to give a general idea but I think a callback (returning true or false based on a component type) here would be more appropriate...

I think this addition is not necessary as we have already [`attributes` option in `toHTML`](https://grapesjs.com/docs/api/component.html#tohtml) (you can also pass a function to change attributes dynamically based on the component)

PR #5017, opened 2023-03-31
fix: resolved bug with top/left being sticky when resizing from l/t side

PaulRill00:
When someone is resizing from the left/top side, the element is sticky to the top/left.
This fix resolves that issue, where it will use opposite sides when resizing components.

https://github.com/GrapesJS/grapesjs/discussions/5015


artf:
Awesome, thanks @PaulRill00 added a fix to prevent top/left if not in standard drag mode

Issue #4503, opened 2022-08-05
BUG: Missing inline styles used by CKEditor when adding raw html components 


gustavohleal:
### GrapesJS version

- [X] I confirm to use the latest version of GrapesJS

### What browser are you using?

Chromium v96

### Reproducible demo link

https://grapesjs.com/demo-newsletter-editor.html

### Describe the bug

**How to reproduce the bug?**
1. Open the developer tools at the console tab
2. Run the code below
3. Select the text added
4. Edit something with CKEditor
5. Select another component

**What is the expected behavior?**
Expected to add inline styles from the spans inside the text component so that the CKEditor understand the text-styling and keep them when opened.

**What is the current behavior?**
Adding the raw HTML by `editor.addComponent()` do add the styling, but once you edit the text and blur it the style is removed from the text segments where they belong. This happens only with CKEditor because it uses the `<span>` and inline style to edit the text.
I'm putting the newsletter preset demo because it is the only demo from grapesjs that use CKEditor.
If is necessary to execute some code in order to reproduce the bug, paste it here below:
```js
editor.DomComponents.clear()
editor.addComponents(`<div data-gjs-type="text"><span style="color: blue;">Teste</span> de <span style="font-family: arial; font-size: 25px"><b>span</b></span> inline</div>`)

```
I need to add components this way so that legacy HTML models and templates can be opened into grapesjs. Sometimes CKEditor even does not recognizes that `spans` or `<b>` elements are text segments and throw an error as it does not work with  those text styling elements.
This problem is similar to #3069 that was fixed and I believe this has returned once the feature of transforming inline styles into classes was added. 
Can we somehow keep the elements inside text components as raw html?

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

shkhalid:
+1

pety-dc:
Hey
I've been having similar issues. I think it's quite serious as both newsletter and mjml demo are affected. (Both use CKeditor as RTE)

When I import a content that has a text (mj-text) that has styled elements (span) the imported content gets parsed and the styling elements are recognized as individual components and not as contents of the wrapping text element.


example (This is valid according to mjml.io):
```
<mjml>
  <mj-body>
    <mj-section>
      <mj-column>
        <mj-image width="100px" src="/assets/img/logo-small.png"></mj-image>
        <mj-divider border-color="#F45E43"></mj-divider>
[...]

PR #5071, opened 2023-04-26
Fix nested text problem with CKEditor by adding disableNestingForTypes config option to parser

pety-dc:
There are some serious problems with using CKEditor as an alternative RTE for grapesJS.

The default behaviour (that can be observed in the webpage demo) is that nested texts are not only allowed, but adding wrapper spans around texts is the way to format parts. The wrapper elements (nested texts) are then handled as individual components and can be formatted with grapesJS tools.

CKEditor also uses spans and other inline html elements with style attribute for formatting. The problem is that during loading such content, grapesjs parses into text content and recognizes the wrapper elements as nested text components and strips style attributes from them. 

detailed description of the problem can be found in issue #4503 

Default text components are not defined by the model's isComponent method, but rather later in the parseNode method at:
```
// If all children are texts and there is some textnode the parent should
// be text too otherwise I'm unable to edit texnodes
```
Therefore the behaviour cannot be controlled in components' isComponent method.

The proposed solution prevents grapesJS from parsing into selected component types for nested components, and adds the node's innerHTML as content instead. 

The proposed configuration option allows the developer to set which types' contents should be handled encapsulated.

### DIFF
--- a/src/parser/model/ParserHtml.ts
+++ b/src/parser/model/ParserHtml.ts
@@ -122,6 +122,7 @@ const ParserHtml = (em?: EditorModel, config: ParserConfig = {}) => {
       const nodes = el.childNodes;
 
       for (var i = 0, len = nodes.length; i < len; i++) {
+        const conf = em?.get('Config') || {};
         const node = nodes[i] as HTMLElement;
         const attrs = node.attributes || [];
         const attrsLen = attrs.length;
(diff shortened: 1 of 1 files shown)


artf:
Thanks @pety-dc I appreciate the effort for the PR. The idea itself is legit but unfortunately I'm not able to merge it as it will allow XSS. I'll try to find a way to fix those issue, if you have a safer alternative I'll be open to review it

PR #5102, opened 2023-05-10
fix: updated some typescript exports

PaulRill00:
Since the latest change, some of the typescript definitions are missing, therefore I re-added some of the typescript exports.
- `Component`
- `ComponentView`
- `Components`
- `ComponentManager`

Besides that, there was an issue with the `Editor#setZoom` requiring a string, whilst actually needing a number or string. Therefore I updated the type for that as well.


dist/grapes.min.js
PaulRill00, line 1: Im not sure if the changes in both `dist/` and `package.json` are required, I can revert these if needed.

atlantamatt: Traits, Trait and TraitView would be helpful as well. Im still going through the exported types, but seems like much more needs be exported: Pages, Page, Frames, Canvas, Dragger, etc.

artf: Thanks @PaulRill00  🙏
Yeah correct, everything inside `dist/` shouldn't be committed in PRs as it's all generated during build/publish (same as version in `package.json`). One day I'll pull that dir off from the repo 😅

I actually pushed some commits containing similar changes (exporting models mainly) but feel free to update with other exports

PaulRill00:
I've opened a new PR in favour of a cleaner commit history here: https://github.com/GrapesJS/grapesjs/pull/5106

PR #5103, opened 2023-05-10
fix: element jumping when resizing on zoomed frame

PaulRill00:
When resizing elements no a zoomed frame `Editor#setZoom(50)` the element will jump around.
This fix will ignore the zoomed context and frame offset, so that the position is correct.

I tested this for absolute both `normal` and `absolute` dragMode.


artf (approved):
LGTM 👍 
Thank you @PaulRill00

PR #5106, opened 2023-05-11
added some more ts exports and fixed Editor#setZoom type

PaulRill00:
Since the latest change, some of the typescript definitions are missing, therefore I re-added some of the typescript exports.
- `ComponentView`
- `ComponentManager`

Besides that, there was an issue with the `Editor#setZoom` requiring a string, whilst actually needing a number or string. Therefore I updated the type for that as well.


artf (approved):

artf:
Thanks @PaulRill00

PR #5114, opened 2023-05-15
update German locale

c9a2334:
Updated German locale + support '-sub' styles


kirill-malyhin:
I'm working with German colleagues, can confirm, that these translations are correct

artf (approved):

artf:
Thanks 🙏

Issue #5175, opened 2023-06-10
BUG: Move method is not working as expected


krishnaeverestengineering:
### GrapesJS version

- [X] I confirm to use the latest version of GrapesJS

### What browser are you using?

Chrome

### Reproducible demo link

https://codesandbox.io/s/adoring-margulis-873kdv

#### Context

Move component method doesn't work as expected.

```
if (component) {
      const { at } = opts;
      const index = this.index();
      const sameParent = component === this.parent();
      const sameIndex = index === at || index === at! - 1;

      if (!sameParent || !sameIndex) {
        if (sameParent && at && at > index) {
          opts.at = at - 1;
        }
        this.remove({ temporary: 1 });
        component.append(this, opts);
        this.emitUpdate();
      }
    }
    return this;
```

This piece of code doesn't work as expected.
```

if (component) {
      const { at } = opts;
      const index = this.index();
      const sameParent = component === this.parent();
      const sameIndex = index === at;

      if (!sameParent || !sameIndex) {
        this.remove({ temporary: 1 });
        component.append(this, opts);
        this.emitUpdate();
      }
    }
    return this;
```

this piece of code should work. When sameParent is false and target index is 1, then it won't move to 0 to 1 position for example

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

krishnaeverestengineering:
Hi @artf. I have fixed this issue. Could you please provide access to the repo. I can raise a PR. Thanks

krishnaeverestengineering:
https://github.com/GrapesJS/grapesjs/pull/5176

artf:
Thanks @krishnaeverestengineering but this is actually the expected behaviour for moving the component inside the same parent. This is because the `at` should account also for the index of the component itself.

```
- index: 0
{ Component 1 }
- index: 1
{ Component 2 }
- index: 2
```
So, if you try to move "Component 1" at index 0 or 1, nothing will happen. If you need to move it below "Component 2" you have to indicate index 2

krishnaeverestengineering:
> Thanks @krishnaeverestengineering but this is actually the expected behaviour for moving the component inside the same parent. This is because the `at` should account also for the index of the component itself.
> 
> ```
> - index: 0
> { Component 1 }
> - index: 1
> { Component 2 }
> - index: 2
> ```
> 
> So, if you try to move "Component 1" at index 0 or 1, nothing will happen. If you need to move it below "Component 2" you have to indicate index 2

[...]

PR #5176, opened 2023-06-12
Fix component move method

krishnaeverestengineering:
Fix Component `move` method is not working as expected

BUG Raised: https://github.com/GrapesJS/grapesjs/issues/5175


chethanEvEng (approved):

artf:
https://github.com/GrapesJS/grapesjs/issues/5175#issuecomment-1592692281

PR #5192, opened 2023-06-17
Improve typings

Julia-Alberici:


src/storage_manager/config/config.ts
```diff
@@ -16,7 +16,7 @@ export interface StorageManagerConfig {
    * Available by default: 'local' | 'remote'
    * @default 'local'
    */
-  type?: string;
+  type?: 'local' | 'remote';
```
artf, line 19: I'll replace this with `LiteralUnion<'local' | 'remote', string>` as the type could be a custom name

artf (approved):

artf:
Thanks @Julia-Alberici

PR #5201, opened 2023-06-25
Update rich_text_editor.md

gxanshu:
The previous code snippet encountered a syntax error when attempting to use it in my application. To ensure its usability for others, I have made necessary updates to resolve the syntax error and enable successful execution. These modifications aim to provide a code snippet that can be easily utilized by anyone without encountering any syntax-related issues.

### DIFF
--- a/docs/api/rich_text_editor.md
+++ b/docs/api/rich_text_editor.md
@@ -65,8 +65,8 @@ rte.add('bold', {
 });
 rte.add('link', {
   icon: document.getElementById('t'),
-  attributes: {title: 'Link',}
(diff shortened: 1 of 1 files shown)


artf (approved):

artf:
Thanks @gxanshu but keep in mind that all .md files in the `api` folder are generated from the source, so the proper file to update would be https://github.com/GrapesJS/grapesjs/blob/dev/src/rich_text_editor/index.ts (I'll update it with your changes)

gxanshu:
Thank you 😊

PR #5358, opened 2023-09-03
resolves #5357

### LINKED ISSUE #5357: BUG: (type issue) cannot pass HTMLElement as icon when adding new rte action
### GrapesJS version

Chrome

### Reproducible demo link

https://jsfiddle.net/vmeywbk8/11/

### Describe the bug

Just a small type issue we came across - when adding an action to the RichTextEditor, the "icon" property of RichTextEditorAction is marked as string.
But in addAction
(https://github.com/GrapesJS/grapesjs/blob/793599cbd5493c0104db1134bd4b1684341274a7/src/rich_text_editor/model/RichTextEditor.ts#L363)
if we pass in a HTMLElement, it will be appended to the button. This still works fine at runtime, but you will get a type mismatch error from TypeScript if you try to pass anything other than a string.

the "icon" property should probably be marked as "string | HTMLElement" to resolve this - I'd like to fix this as my first contribution

### Code of Conduct

peters-ben-0007:
updating the type of RichTextEditorAction.icon to be "string | HTMLElement"

### DIFF
--- a/src/rich_text_editor/model/RichTextEditor.ts
+++ b/src/rich_text_editor/model/RichTextEditor.ts
@@ -9,7 +9,7 @@ import { getModel } from '../../utils/mixins';
 
 export interface RichTextEditorAction {
   name: string;
-  icon: string;
+  icon: string | HTMLElement;
   event?: string;
   attributes?: Record<string, any>;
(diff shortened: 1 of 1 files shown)


artf (approved):

artf:
Thanks @peters-ben-0007

PR #5437, opened 2023-10-02
Add CSS variables

quentin-bettoum:
The goal of this PR is to add CSS variables as an alternative to SASS variables to customize GrapesJS.

I had to replace the use of SASS `lighten` and `darken` as they cannot work with CSS variable since they are interpreted at compile-time.
I used [color-mix](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix) instead, but the percentage used in `lighten`/`darken` and in `color-mix` doesn't give the same result as [explained here](https://frzi.medium.com/less-sass-more-color-mix-or-color-manipulation-with-pure-css-dbf6f244ef29). From what I've seen, it doesn't seem to make a big difference, but we have to keep this in mind.

I also took the opportunity to remove some SCSS variables that were not used anymore.


artf (approved):

artf:
Awesome work @quentin-bettoum 🔥
I think I'll probably switch `--grapes-` to `--gjs-` but overall looks all good 🙇‍♂️

PR #5508, opened 2023-11-08
Type argument in replaceWith in Component.ts

oliviermonnier:


artf:
Just merged the same fix [here](https://github.com/GrapesJS/grapesjs/pull/5504) 😅

PR #5546, opened 2023-12-05
Fix Portuguese translations

brenoassp:
I noticed that some Portuguese translations were not working because the properties' names changed but they were not updated in the Portuguese translation file.

I updated the properties' names and also fixed the translation to be consistent with the names in the English translation file.


artf (approved):

artf:
Thank you @brenoassp

PR #5554, opened 2023-12-13
Update lint dependencies

xQwexx:
I got the following error when I wanted to commit some changes:

```
✖ eslint --ext .ts,.js --fix found some errors. Please fix them and try committing again.

/Users/alexrt/Documents/Work/grapesjs/src/commands/index.ts
  0:0  error  Parsing error: DeprecationError: 'originalKeywordKind' has been deprecated since v5.0.0 and can no longer be used. Use 'identifierToKeywordKind(identifier)' instead
```
  
 The solution was updating the lint dependencies


package.json
```diff
@@ -41,17 +41,17 @@
...
     "@vuepress/plugin-google-analytics": "^1.8.2",
     "documentation": "^13.2.5",
-    "eslint": "^8.12.0",
+    "eslint": "^8.55.0",
     "grapesjs-cli": "^4.1.3",
     "html-entities": "^1.4.0",
     "husky": "^2.7.0",
     "jest": "^24.9.0",
-    "lint-staged": "^8.2.1",
+    "lint-staged": "^15.2.0",
```
artf, line 54: Do we need to update lint-staged? Looks like quite a big jump in the major versions, what are the breaking changes here?

xQwexx: 15.0.0
BREAKING CHANGES
Require at least Node.js 18.12.0
This release drops support for Node.js 16, which is EOL after 2023-09-11. Please upgrade your Node.js to the latest version.

[v14.0.0](https://github.com/okonet/lint-staged/releases/tag/v14.0.0) - 13 Aug 2023
BREAKING CHANGES
Please upgrade your Node.js version to at least 16.14.0.

[v13.0.0](https://github.com/okonet/lint-staged/releases/tag/v13.0.0) - 01 Jun 2022
BREAKING CHANGES
lint-staged will no longer support Node.js 12, which is EOL since 30 April 2022

[v12.0.0](https://github.com/okonet/lint-staged/releases/tag/v12.0.0) - 13 Nov 2021
BREAKING CHANGES
lint-staged is now a pure ESM module, and thus requires Node.js version ^12.20.0 || ^14.13.1 || >=16.0.0.
To update your Node.js integration, please use:
```
// const lintStaged = require('lint-staged')
import lintStaged from 'lint-staged'
```

[v11.0.0](https://github.com/okonet/lint-staged/releases/tag/v11.0.0) - 07 May 2021
BREAKING CHANGES
Node.js 12 LTS 'Erbium' is now the minimum required version

[v10.0.0](https://github.com/okonet/lint-staged/releases/tag/v10.0.0) - 19 Jan 2020
BREAKING CHANGES
Previously, lint-staged would allow empty commits in the situation where a linter task like "prettier --write" reverts all staged changes automatically. Now the default behaviour is to throw an error with a helpful warning message. The --allow empty option can be used to allow empty commits, or allowEmpty: true for the Node.js API.
Node.js v8 is no longer supported because it will reach EOL on 2019-12-31
Prior to version 10, tasks had to manually include git add as the final step. This behavior has been integrated into lint-staged itself in order to prevent race conditions with multiple tasks editing the same files. If lint-staged detects git add in task configurations, it will show a warning in the console. Please remove git add from your configuration after upgrading.

[v9.0.0](https://github.com/okonet/lint-staged/releases/tag/v9.0.0) - 01 Jul 2019
BREAKING CHANGES
The advanced configuration options have been deprecated in favour of the simple format
Local commands are no longer resolved by lint-staged, but execa will do this instead. In effect, there are no longer pretty error messages when commands are not found.
Very long arguments strings are no longer chunked on Windows. Function linters should be used instead to customise this behaviour.

xQwexx: I listed out the breaking changes I could downgrade to the version that you think makes sense.

danstarns:
@xQwexx We are going to close this in favour of: 

* https://github.com/GrapesJS/grapesjs/issues/6045

Thank you for your efforts :)

PR #5559, opened 2023-12-14
Support rotating of elements

PaulRill00:
This PR introduces an option to use a rotate handle to rotate elements.

```
component: {
  rotatable: true
}
```

Next to adding support for rotating, this PR also adds a fix for resizing rotated elements. The element will now 'lock' on the opposite handle when someone stats resizing.

_In this PR CSS changes have been made, however, I don't know if we should submit the minified CSS in this PR._


artf:
Thanks @PaulRill00 this is a nice functionality but I'm against merging it in the core, especially if no one ever asked for it... I'm quite confused as to why you even started this without opening a discussion.
If the next time you wish to add something, please let's discuss this first.

Anyway, I'd suggest you refactor this to a plugin, I see this functionality as a perfect usage of [Canvas Spots](https://grapesjs.com/docs/modules/Canvas.html#canvas-spots)

PaulRill00:
> Thanks @PaulRill00 this is a nice functionality but I'm against merging it in the core, especially if no one ever asked for it... I'm quite confused as to why you even started this without opening a discussion. If the next time you wish to add something, please let's discuss this first.
> 
> Anyway, I'd suggest you refactor this to a plugin, I see this functionality as a perfect usage of [Canvas Spots](https://grapesjs.com/docs/modules/Canvas.html#canvas-spots)

Hi @artf, I would argue that at least half of the functionality included in this PR should be included in the core. Especially since the 'absolute' mode is already supported. But - when you set any rotation on the element using the build-in editor, the resize/drag functionality is broken.

Then - it only seems like a small addition to also natively support a rotate-handle to make navigating elements easier.

As to the reason why I never opened a discussion, is because we use GrapesJS in a project internally and were missing this functionality. So we decided to add it ourselves and then open a PR here.

Moving this to a plugin sounds like a lot of work. Maybe you could re-consider implementing this into the core of GrapesJS? We like using the project and would love to make more contributions, but not merging this would mean we'll be using a fork - and therefore a detached-HEAD.

artf:
Sorry @PaulRill00 but in my opinion, just the Rotator file is a HUGE addition to the core, that a lot of people will download and probably won't even use, so for me it's a hard pass for this addition to the core.

> Moving this to a plugin sounds like a lot of work

I'm not sure about that, what you might find it's some kind of missing API or event that would you allow to bind easily your functionality but here I'm more than open to implementing it if necessary.

PaulRill00:
@artf - we're thinking about finally turning this into a proper plugin.
But for that we are wondering if our additions towards the `Resizer` would be OK to be merged. Since in our opinion it's weird that the current resize logic does not work correctly when the element is rotated.
Then we will move the `Rotator` logic + handles etc to a plugin for easy injection.

artf:
@PaulRill00 can you please open a new issue/PR showing the problem (reproducible demo) with the Resizer.

lukjaki:
@PaulRill00 is this available as a plugin?

Issue #2282, opened 2019-09-19
[Feature Proposal] hyperhtml as a replacement for virtual dom


pouyamiralayi:
Hi there! the currently used structure of grapesjs is based on leveraging _backbonejs model-view concept_ for implementing the **Virtual Dom** inside the ecosystem of grapesjs. Virtual Dom is a great practice which has a vast area of applications and benefits; but the **downside** of virtual dom is the _actual gap between virtual and actual dom itself_, what this mean is that the usage of virtual dom is some sort of **performance sacrifice** in order to achieve the model-view concept which is the already established architecture of grapesjs, react, vue & etc.
based on the real-time nature of grapesjs and its ability to manage thousands of model objects based on each user interaction, there are cases that performance is poorly appreciated on model updates that as it's result there would be no real-time experience for the end user. as an example of this scenario think about calling **addAttributes** for _property changes on a complex component type_ that will be invoked every time the end user provides us with new inputs in the associated trait section; each time that user presses a key, there would be a noticeable delay for the model to updates the attributes of the target component according to the user input.
**my proposal on this matter** is to get rid of virtual dom in the sack of **hyperhtml** that benefits from actual dom that are _bounded to a context object that acts as our current model_. the way hyperhtml is achieving the model-view concept for actual dom is leveraging js native **template literals** and it's _associated chunks and interpolations_.  i know this is a very scary taboo for the current implementation :) but i think if grapesjs architecture has been started up with backbone, the sub parts of the ecosystem should not follow the same lead although this is a common implication in the development path of a complex project like grapesjs; _backbone can be really tempting to push developers mind to go with it until the end_, but i am suggesting more flexibility by separating grapesjs into a s**olid core with backbone** and all the other parts leveraging hyperhtml to introduce a more lightweight, performance friendly ecosystem that embraces the future with _no fear of performance breakdowns.
i am a very naive developer and please forgive me if i am going wrong with this.
cheers.

lexoyo:
Nice idea
I use lit-html quite lot, it's the same approach but also adds web components when/if needed
But i don't get where would backbone stop? And what would be without it. I guess it is a question that is broader about what @artf announced in the roadmap (taking things out of the core)

artf:
Yes @lexoyo the final goal here would be to stop relying on Backbone's View and migrate all editor UI elements to web components. In the end, there will be `@grapesjs/core` (no UI) and `@grapesjs/core-ui` with a set of reusable and extendable web components for the editor UI

[...]

PR #5562, opened 2023-12-16
Trait refactor

xQwexx:
Hi @artf
I am working of refactoring of the Trait module because I wanted to build some custom input layouts but there was not existing generic ones that I could use, so I started to work on a generic InputViews (Backbone model attribute modifiers),
this would also enable to use it instead of PropertyView, TraitView and InputView so it will be much easier to maintain and understand

To a more reusable which would not depend on the component and trait to be a backbone model:

But I think this would require to break the current API for the Traits in some places, so I wanted to know what do you thinks of this before I do a clean pr (so this is a WIP branch of mine with some unrelated changes)

```
      let input = new TraitTextView(em, { name: 'route' })
      input.setTarget('name', this.model, { changeProp: true });
      this.$el.append(input.render().el);
```


artf:
Yeah I'd stay away right now from any kind of breaking change. Traits could be definitely improved and having one generic Trait model like at the moment is for sure not ideal. 

If I had to refactor Traits now, I'd make it more similar to how StyleManager properties are managed, where I have one model and view per type, which also makes more sense from a basic OO point of view.

Agree about having a hard time right now reusing core UI components in GrapesJS, indeed I'd like to ditch completely from the Backbone's View and create a [package of reusable web components](https://github.com/GrapesJS/grapesjs/issues/2282#issuecomment-1845683657). That would also allow plugin developers to reuse easily the same library of UI components.
That's also the reason why I'd like to reduce to a minimum any kind of new UI inside GrapesJS right now (and there is plenty in your PR unfortunately)

artf:
Hey @xQwexx I'm closing this as I'm not planning to merge it and it generates noise.
Let me know if you're open to discuss how to make the editor UI better with reusable web components as mentioned in a previous post.

PR #5565, opened 2023-12-18
Fix component script param parse if its a function 

xQwexx:
I tried to add script param for a component but it was having an error
```
const script = ()=>{
  const el: HTMLDivElement = this as any;
  console.log(el.id)
  // this is bound to the component element
  console.log('the element', this);
};
```
because it was converted to:
```
ƒ () {
    var el = _this;
    console.log(el.id);
    // this is bound to the component element
    console.log('the element', _this);
}
```


artf:
I think there is more of an issue with the compiler as you're using `this` and an array function (so the bundler tries to keep `this` context for you).
I'd simply avoid using array functions with component scripts as the context is never the same and that might bring similar issues.
Anyway, replacing blindly `replaceAll('_this', 'this')` will never be a solid fix as you risk replacing something unintentionally (considering in JS you can do a lot of dynamic stuff)

Issue #5621, opened 2024-01-15
BUG: Border is not taken into consideration when showing placeholder on empty elements


ronaldohoch:
### GrapesJS version

- [X] I confirm to use the latest version of GrapesJS

### What browser are you using?

Edge last version

### Reproducible demo link

https://grapesjs.com/demo

### Describe the bug

**How to reproduce the bug?**
1. Get an empty element
2. Add double border with high value. (20..30px...)
3. Move any block into it.
4. 
![image](https://github.com/GrapesJS/grapesjs/assets/2287371/5894b9b2-08c1-46f9-a62f-536e221839d1)

**What is the expected behavior?**
Placeholder should be placed in center of element.

**What is the current behavior?**
Placeholder is shown in the top of element, inside boder

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #5622, opened 2024-01-15
Fix: #5621

ronaldohoch:


artf:
Thanks for the input @ronaldohoch did the same [fix](https://github.com/GrapesJS/grapesjs/commit/ec4f2d634e28bf8b9da75880ed51d5b4beed44ac) but took also into account the possibility of individual borders (eg. `border` would be empty in case of `border-top: 2rem solid; border-left: 5rem solid;`)

PR #5642, opened 2024-01-24
docs: add new extensions

bgrand-ch:


artf (approved):

artf:
Nice, thanks @bgrand-ch don't forget to add `grapesjs-plugin` tag/topic to your GH repos so they will be visible from the ["Community Plugins" link](https://github.com/GrapesJS/grapesjs?tab=readme-ov-file#plugins)

Issue #5650, opened 2024-01-29
BUG: When inserting text when component not synced, the text is overwritten from the existing model


Wayne-Mather:
### GrapesJS version

- [X] I confirm to use the latest version of GrapesJS

### What browser are you using?

Edge 121.0.2277.83, Firefox 122

### Reproducible demo link

https://jsfiddle.net/50mga1t7/1/

### Describe the bug

**How to reproduce the bug?**
1. Create a toolbar action to insert some text
2. Edit a text node and insert some text
3. Click the toolbar command to insert some new component/text
4. The textnode will reset to the model details before the edit's took place

**What is the expected behavior?**
The textnode should show the current text as being edited and the new inserted text

**What is the current behavior?**
The new text is inserted using the text from the model rather than the current view

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

artf:
This is not a bug, merging the current DOM with components requires a bit more effort. 
There is already a [similar API](https://github.com/GrapesJS/grapesjs/blob/fb5175fd1edd13a2db54d661193fb61a354f8625/src/rich_text_editor/model/RichTextEditor.ts#L414) you can use or re-implement with your own needs (using component definition instead of HTML string)

PR #5651, opened 2024-01-29
Bugfix/5650 inserting text uses model data not current view

Wayne-Mather:
This is my proposed solution for the bug 5650 when inserting text into a text node that has not synchronised back to the model yet.

[BUG: When inserting text when component not synced, the text is overwritten from the existing model](https://github.com/GrapesJS/grapesjs/issues/5650)


src/dom_components/view/ComponentTextView.ts
```diff
@@ -182,13 +182,13 @@ export default class ComponentTextView extends ComponentView<ComponentText> {
       const offset = range.startOffset;
       const textModel = getComponentModel(textNode);
       const newCmps: (ComponentDefinition | Component)[] = [];
-
+      let data = textNode.textContent ?? '';
```
bgrand-ch, line 185: suggestion: `const data = textNode.textContent || '';`, because the `data` is not updated afterwards.

src/dom_components/view/ComponentTextView.ts
```diff
@@ -182,13 +182,13 @@ export default class ComponentTextView extends ComponentView<ComponentText> {
...
       if (textModel && textModel.is?.('textnode')) {
         const cmps = textModel.collection;
         cmps.forEach(cmp => {
           if (cmp === textModel) {
             const type = 'textnode';
-            const cnt = cmp.content;
+            const cnt = (data = '' ? cmp.content : data);
```
bgrand-ch, line 191: important: use equality operator (`==`) or strict equality operator (`===`)

suggestion: `!data ? cmp.content : data;` (grouping operator `(...)` is not necessary here)

bgrand-ch (changes requested):

src/dom_components/view/ComponentTextView.ts
Wayne-Mather: done

src/dom_components/view/ComponentTextView.ts
Wayne-Mather: done

bgrand-ch (approved):

artf:
This might lead to unexpected behavior as it's not always the case I want to use the DOM content. 

I'd accept this only behind an additional option flag (eg. `textView.insertComponent({...}, { useDomContent: true })`)

Wayne-Mather:
@artf I have wrapped the code to test if that property exists and if it's true. That way content is only replaced if insertComponent(xx, { useDomContent: true }) is used

artf (approved):

artf:
Thanks @Wayne-Mather

PR #5704, opened 2024-02-25
ISSUE 5703: Color picker offset relative to editor container

bernesto:
The picker now opens relative to the editor container and not the window.


artf (approved):

artf:
Thanks @bernesto 🙇‍♂️

PR #5706, opened 2024-02-26
Move out Script generation logic from the Component

xQwexx:
I think the Component contains to much logic and it is hard to comprehend all of its features,
I would want to add some other functionality for the script generation part but it happen in multiple places the component, JsGenerator and the CanvasView I would like to move all of it into the ScriptSubComponent first I just moved from the Component so I could do the changes incrementally.


ShivamJha2436 (approved):

danstarns:
Thank you @xQwexx, let's now close this as it's gone stale. We appreciate you!

PR #5736, opened 2024-03-09
Always place the tools above the highlighter

bernesto:
Component highlight indicators display above RTE tools on hover. This sets the z-index 1 above the indicators.

Current:
image

Proposed:
image


artf (approved):
Makes sense 👍

PR #5750, opened 2024-03-13
Update InputColor to append at the body instead of the editor

bernesto:
The addresses an issue where we are trying to place a popover palette absolutely positioned over a InputColor element at an arbitrary location on the page. And we need to do this even when the origin element is outside of the editor container node.

To meet this requirement, the popover palette can not be contained anywhere within an offsetParent, and further it must be at the highest z-index. Otherwise it will be subject to the visible area of it's offsetParent and ordered by the parent's z-index. So we are adding that container directly to the `body`.


artf:
Thank you @bernesto Nice job, I really appreciate your prompt response.

PR #5815, opened 2024-04-17
Up doc.write to element.innerHTML

WooRung:
remove warning on recent chrome


artf (changes requested):
It's not the same thing. With this change the default `<!DOCTYPE html>` won't be rendered and that changes the way the iframe is rendering the content.

Screenshot 2024-04-19 at 19 02 58

PR #5847, opened 2024-04-29
Move last selected component

stfnzl:
In absolute mode, there is a discrepancy between the toolbar, which is shown on the latest selected element, and the moving command, which is triggered on the first selected one.

**Before**

before

**After**

after


artf (approved):
👍

Issue #5525, opened 2023-11-19
BUG: Performance issue when adding more than 20 classes to a single component


duydvu:
### GrapesJS version

- [X] I confirm to use the latest version of GrapesJS

### What browser are you using?

Edge v119.0.2151.72

### Reproducible demo link

https://grapesjs.com/demo.html

### Describe the bug

Hi, thank you for creating an amazing project. I would like to report an issue found during my use of GrapesJS.

**How to reproduce the bug?**
1. Select a component
2. Add classes with arbitrary names until the component has over 20 classes.
3. Select another component and re-select the component.

**What is the expected behavior?**
There should be no performance issue.

**What is the current behavior?**
The browser freezes for a noticeable amount of time. The more classes added, the longer it freezes. My browser freezes for 10 seconds when the component has 23 classes.

This is a common issue when working with TailwindCSS since there are a lot of classes. For example, if I copy an HTML template from a TailwindCSS library, the browser might freeze when I import to the Custom Code.

**Cause**
So I investigated and found the problem. Here it is:

![image](https://github.com/GrapesJS/grapesjs/assets/23261928/d332aa8f-546e-4b08-85ea-9cc0e184e8d2)

The `combine` function generates all combinations of the classes, which results in 2^n combinations (n is the number of classes). So if there are 20 classes, there would be 1048576 combinations.

If I understand it right, it generates all combinations to filter the CSS rules of the component in this function: https://github.com/GrapesJS/grapesjs/blob/8025b95a645907574389e950f32950190ceb602c/src/css_composer/index.ts#L373

![image](https://github.com/GrapesJS/grapesjs/assets/23261928/69072d86-f3e3-4b24-ba06-26a08065437c)

I think this algorithm is not optimal since it loops through all selectors and rules. A better solution is only looping through all the rules and for each rule, checking if the rule's selectors is a subset of the component's selectors.

Simplified example:
Component's selectors: `['.a', '.b', '.c', '.d', '.e', '.f']`
Rules' selectors:
```js
[
  ['.a', '.b', '.c'],
  ['.x', '.y', '.z'],
]
```
For each rule, you only need to check if `['.a', '.b', '.c']` is a subset of `['.a', '.b', '.c', '.d', '.e', '.f']`, without generating all combinations of 6 classes.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

artf:
Agree with the suggested alternative, I will update accordingly. Thanks @duydvu

PR #5890, opened 2024-05-17
Show style applied on tag (p, span, div) and private selectors as parent rules instead of hiding them

Dobby85:
### Current behavior

The current behavior describe below can be tested on this JSFiddle : https://jsfiddle.net/728urdh3/

#### Context

I add a custom component (`sectionBlock`) that have styles by the `section` tag name and the private selector `ed-layout-div`.

```css
section {
  width: 940px;
  min-height: 50px;
  border: 1px dashed black;
}

.ed-layout-div {
  margin-top: 20px;
  margin-bottom: 30px;          
}
```

#### Step to reproduce

When you drop this component in the canvas and check it's style.

- You **do not see** `width` and `min-height` properties set on the section.
- You **see** the `margin-top` and the `margin-bottom` apply on the hidden selectors.

Screenshot 2024-05-17 at 08 25 43

Now when you add a class to the component :

- You **do not see anymore** the `margin-top` and the `margin-bottom` apply on the hidden selectors.

Screenshot 2024-05-17 at 08 27 20

#### Context

The context is the same than the previous one

#### Step to reproduce

When you drop this component in the canvas and check it's style.

- You **see** `width` and `min-height` properties set on the section as parent rules
- You **see** `margin-top` and `margin-bottom` apply on the hidden selectors as parent rules

Screenshot 2024-05-17 at 08 28 16

Now when you add a class to the component :

- You still **see** `width` and `min-height` properties set on the section as parent rules
- You still **see** `margin-top` and `margin-bottom` apply on the hidden selectors as parent rules

Screenshot 2024-05-17 at 08 28 31

### Why do I do that?

I think it's more comfortable to always see the style that is apply on the component to better understand why it is like it is. You still can override style but you always know which rule are applied to the component.

All tests passed and I add 3 new tests to test style apply on tag name and return style with multiple classes.

Do not hesitate if you need more information or if I have to update/fix something !

Have a great day!

### DIFF
--- a/src/style_manager/index.ts
+++ b/src/style_manager/index.ts
@@ -81,7 +81,8 @@ import StyleableModel, { StyleProps } from '../domain_abstract/model/StyleableMo
 import { CustomPropertyView } from './view/PropertyView';
 import { PropertySelectProps } from './model/PropertySelect';
 import { PropertyNumberProps } from './model/PropertyNumber';
-import { PropertyStackProps } from './model/PropertyStack';
+import PropertyStack, { PropertyStackProps } from './model/PropertyStack';
+import PropertyComposite from './model/PropertyComposite';
 
 export type PropertyTypes = PropertyStackProps | PropertySelectProps | PropertyNumberProps;
 
@@ -309,7 +310,7 @@ export default class StyleManager extends ItemManagerModule<
    */
   addProperty(sectorId: string, property: PropertyTypes, opts: AddOptions = {}): Property | undefined {
     const sector = this.getSector(sectorId, { warn: true });
-    let prop = null;
+    let prop;
     if (sector) prop = sector.addProperty(property, opts);
 
     return prop;
@@ -522,7 +523,7 @@ export default class StyleManager extends ItemManagerModule<
    *  options: [{ id: 'value1', label: 'Some label' }, ...],
    * })
    */
-  addBuiltIn(prop: string, definition: Omit<PropertyProps, 'property'> & { proeperty?: 'string' }) {
+  addBuiltIn(prop: string, definition: PropertyProps) {
     return this.builtIn.add(prop, definition);
   }
 
@@ -597,6 +598,7 @@ export default class StyleManager extends ItemManagerModule<
       const optsSel = { array: true } as const;
       let cmpRules: CssRule[] = [];
       let tagNameRules: CssRule[] = [];
+      let invisibleRules: CssRule[] = [];
       let otherRules: CssRule[] = [];
       let rules: CssRule[] = [];
 
@@ -641,8 +643,19 @@ export default class StyleManager extends ItemManagerModule<
       } else {
         cmpRules = sel ? cssC.getRules(`#${sel.getId()}`) : [];
         tagNameRules = rulesByTagName(sel?.get('tagName') || '');
+        // Get rules set on active and visible selectors
         otherRules = rulesBySelectors(target.getSelectors().getFullName(optsSel));
-        rules = tagNameRules.concat(cmpRules).concat(otherRules);
+        // Get rules set on invisible selectors like private one
+        const allCmpClasses = sel?.getSelectors().getFullName(optsSel) || [];
+        const invisibleSel = allCmpClasses.filter(
+          (item: string) =>
+            target
+              .getSelectors()
+              .getFullName(optsSel)
+              .findIndex(sel => sel === item) === -1
+        );
+        invisibleRules = rulesBySelectors(invisibleSel);
+        rules = tagNameRules.concat(cmpRules).concat(invisibleRules).concat(otherRules);
       }
 
       const all = rules
@@ -813,7 +826,7 @@ export default class StyleManager extends ItemManagerModule<
     });
   }
 
-  __upProp(prop: any, style: StyleProps, parentStyles: any[], opts: any) {
+  __upProp(prop: Property, style: StyleProps, parentStyles: any[], opts: any) {
     const name = prop.getName();
     const value = style[name];
     const hasVal = propDef(value);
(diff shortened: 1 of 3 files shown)


src/style_manager/index.ts
```diff
@@ -597,27 +597,65 @@ export default class StyleManager extends ItemManagerModule<
...
+          const rSels = rule.getSelectors().map(s => s.getFullName());
+
+          // rSels is equal to 0 when rule selectors contain tagName like : p {}, .logo path {}, ul li {}
+          if (rSels.length === 0) {
+            return false;
+          }
+
+          return rSels.every(rSel => values.indexOf(rSel) >= 0);
+        });
+      };
```
artf, line 616: What is the point of this change and why did you remove `!values.length ? []`? That was important to avoid performance issues. 
We don't want to loop over all the rules if `values` is empty

src/style_manager/index.ts
```diff
@@ -597,27 +597,65 @@ export default class StyleManager extends ItemManagerModule<
...
+
+      const rulesByTagName = (tagName: string) => {
+        return cssC.getRules().filter(rule => {
+          const ruleTags = rule
+            .selectorsToString()
+            .split(' ')
+            .filter(item => {
+              return item.startsWith('.') === false && item.startsWith('#') === false;
```
artf, line 624: why not to simply check `rule.selectorsToString() === tagName`?
The logic here will take wrong rules, eg. `.something DIV .something-else`

src/style_manager/index.ts
```diff
@@ -597,27 +597,65 @@ export default class StyleManager extends ItemManagerModule<
...
       // Componente related rule
       if (cmp) {
         cmpRules = cssC.getRules(`#${cmp.getId()}`);
+        tagNameRules = rulesByTagName(cmp.get('tagName'));
         otherRules = sel ? rulesBySelectors(sel.getSelectors().getFullName(optsSel)) : [];
-        rules = otherRules.concat(cmpRules);
+        rules = otherRules.concat(tagNameRules).concat(cmpRules);
       } else {
         cmpRules = sel ? cssC.getRules(`#${sel.getId()}`) : [];
+        tagNameRules = rulesByTagName(sel?.get('tagName') || '');
```
artf, line 645: Keep in mind to avoid useless loops when it's possible

src/style_manager/index.ts
```diff
@@ -597,27 +597,65 @@ export default class StyleManager extends ItemManagerModule<
...
+        // Get rules set on invisible selectors like private one
+        const allCmpClasses = sel?.getSelectors().getFullName(optsSel) || [];
+        const invisibleSel = allCmpClasses.filter(
+          (item: string) =>
+            target
+              .getSelectors()
+              .getFullName(optsSel)
+              .findIndex(sel => sel === item) === -1
+        );
+        invisibleRules = rulesBySelectors(invisibleSel);
```
artf, line 657: `rulesBySelectors` is already executed above, it would make sense to combine CssRule selectors + Component ones and pass them to `rulesBySelectors` together

test/specs/style_manager/index.ts
```diff
@@ -195,6 +195,44 @@ describe('StyleManager', () => {
...
+        const [rule1, rule2, rule3] = cssc.addRules(`
+          .cls { color: red; }
+          div { color: yellow; }
+          #id-test { color: blue; }
+        `);
+        em.setSelected(cmp);
+        obj.__upSel();
+        expect(obj.getSelected()).toBe(rule1);
+        expect(obj.getSelectedParents()).toEqual([rule3, rule2]);
+      });
```
artf, line 234: Thanks 🙇‍♂️ it's good to cover with tests this kind of logic

artf (changes requested):
Thank you @Dobby85 that's a nice idea and overall a good PR.
I'd just ask you to review my comments and keep performance in mind here to avoid falling into [similar issues](https://github.com/GrapesJS/grapesjs/issues/5525).

src/style_manager/index.ts
Dobby85: My bad, I don't why I remove it, I just add it again!

src/style_manager/index.ts
Dobby85: I cannot just check `rule.selectorsToString() === tagName` because `tagName` only include the tag name like `div`, `section`, `li`... and not all the path like `rule.selectorsToString()` can have (`.something .something-else div`).

In fact, it's a problem for this case: `.something DIV .something-else`. So I add a check to be sure the tagName is the last element in the `rule.selectorsToString()`.

I know this function is not perfect, it still contains this caveat:

```html
<section class="cls1">
  <div>Text</div>
</section>
```
```css
.cls1 div {
  padding: 10px;
}
```

In the previous case,`.cls1 div` will be return as a rule for both div because I don't how to check a component is a child or not of a component having `.cls1`. Or if we can check that I think we will add a lot of loops...

I don't know what is the best things to do:

1 - Know the caveat and let it like this
2 - Try to fix it and add a lot of checks and loops

In every case, I fix the following case `.something DIV .something-else` and add a test on it :)

src/style_manager/index.ts
Dobby85: Of course, I update code to call `rulesBySelectors` only once!

test/specs/style_manager/index.ts
Dobby85: It's always better when it's tested 😉

Dobby85:
@artf I don't know who is supposed to resolve conversation. I fix everything I could based on your comment, I let you review changes and resolve conversation if it sounds good to you ☺️

src/style_manager/index.ts
artf: I'd suggest to keep it simple for now, only selectors with one and single tagName are accepted, eg. `div` is ok, `.cls1 div` is not.  So let's just do `rule.selectorsToString() === tagName`.

You're not considering a lot of CSS selectors, like in your case `.cls2 div` will be taken as valid, also `.cls2 div div`... it's better to skip those rules instead of showing the wrong ones. 
We won't be able to be precise with the current approach anyway.

src/style_manager/index.ts
artf: Let's place a check 
```js
const rulesByTagName = (tagName: string) => {
        return !tagName ? [] : cssC.getRules().filter(rule => {
        ...
```

Dobby85: I add the check 👍

Dobby85:
All your comments should be resolve @artf. Tell me if it's not! Else I think everything is ready for merge 😁

artf (approved):

artf:
Nice, thanks @Dobby85

src/style_manager/index.ts
chen-can: The 0.17 version is supported.

chen-can:
> 您的所有意见都应该得到解决@artf。如果不是的话请告诉我！不然我想一切都已经准备好进行合并了😁

 <style>
        .mw-package-price,
        .mw-package-price span.price {
          color: red;
        }

        .mw-package-price span {
          color: #4d4d4d;
        }
      </style>
      <p class="mw-package-price">
        <span class="regular-price">$98.00</span>
        <span class="price" id="prod-1-price">$49.00</span>
      </p>
      fail

chen-can:
Version 0.17.29 works and matches correctly, but the newer version does not. I need to use the newer version because grapesjs-style-bg requires a better version

PR #5915, opened 2024-05-29
Working docker image 

labKnowledge:
Now, anyone can either pull the package and compose the Docker, or pull the project in any environment that supports Docker.


artf:
This a client-side-only framework, having a Docker file here makes no sense to me.

PR #5933, opened 2024-06-08
Add an option to return inline style even if we want to avoid it at the beginning

Dobby85:
Hi 👋

Here is the context why I'm creating this PR. I saw that the config parameter `avoidInlineStyle` is now deprecated because of responsive and state management. But in some case we could still need inline style.

For example, if you get a dynamic list with multiple elements and each element has a specific color that you want to use as text color. You can only do it with inline style. But if here we set inline style, this is because you want it and you know you will cannot handle responsive and state on the text color.

### `avoidInlineStyle = true`

If I understand correctly, `avoidInlineStyle` is set to true by default to handle responsive and state management. We can still add inline style with `setStyle('color: red;', { inline: true })` but when we get the final HTML with `toHTML()`, the inline style is remove and style is lost.

Screenshot 2024-06-08 at 09 32 25

### `avoidInlineStyle = false`

When we put `avoidInlineStyle` to false, and we do `setStyle('color: red;', { inline: true })`, the inline style is set correctly but at the CSS generation (`editor.getCss()`), the same style is also added in the CSS code. What we don't want because we explicitly specify `{ inline: true }` when setting the style.

Screenshot 2024-06-08 at 09 29 59

### What do I change?

So I just add one export parameter in `ToHTMLOptions` named `keepInlineStyle` to allow inline style when it is explicitly wanted.

By doing that, I saw that the test suite `Component` was running an editor with `avoidInlineStyle` set to false, which is not the default way of working now if I understand. So I set it to true in the test suite and fix 3 tests that work a little but differently.

### To conclude

```js
// Set style on #id or .class
cmp.setStyle({ color: 'red' })
// OR
cmp.setAttributes({ style: 'color: red' })

// Style has not been set as an attributes but as a CSS rule
console.log(cmp.getAttributes()) // {}
console.log(cmp.getStyle()) // { color: 'red' }


// Set style on #id or .class
cmp.setStyle({ color: 'red' },  { inline: true })
// OR
cmp.setAttributes({ style: 'color: red' }, { inline: true })

// Style has not been set as an attributes but as a CSS rule
console.log(cmp.getAttributes()) // { style: 'color: red;' }
console.log(cmp.getStyle()) // {}
console.log(cmp.getStyle({ inline: true })) // { color: 'red' }
```

I hope my PR is clear and you will understand why I did this.

PS: Something totally different, I add the `keymaps` config key in the editor config type to avoid typescript check issue when using it.

Do not hesitate if you have any reviews or comments!

### DIFF
--- a/src/common/index.ts
+++ b/src/common/index.ts
@@ -5,7 +5,7 @@ interface NOOP {}
 
 export type Debounced = Function & { cancel(): void };
 
-export type SetOptions = Backbone.ModelSetOptions & { avoidStore?: boolean };
+export type SetOptions = Backbone.ModelSetOptions & { avoidStore?: boolean; inline?: boolean };
 
 export type AddOptions = Backbone.AddOptions & { temporary?: boolean };
 
--- a/src/editor/config/config.ts
+++ b/src/editor/config/config.ts
@@ -7,6 +7,7 @@ import { DeviceManagerConfig } from '../../device_manager/config/config';
 import { I18nConfig } from '../../i18n/config';
 import { ModalConfig } from '../../modal_dialog/config/config';
 import { LayerManagerConfig } from '../../navigator/config/config';
+import { KeymapsConfig } from '../../keymaps/config';
 import { PageManagerConfig } from '../../pages/types';
 import { PanelsConfig } from '../../panels/config/config';
 import { ParserConfig } from '../../parser/config/config';
@@ -355,6 +356,11 @@ export interface EditorConfig {
    * Configurations for Commands.
    */
   commands?: CommandsConfig;
+
+  /**
+   * Configurations for keymaps
+   */
+  keymaps?: KeymapsConfig;
 
   /**
    * Configurations for Css Composer.
--- a/test/specs/dom_components/model/Component.ts
+++ b/test/specs/dom_components/model/Component.ts
@@ -280,11 +280,11 @@ describe('Component', () => {
       class: 'class1 class2',
       style: 'color: white; background: #fff',
     });
+    // Style is not in attributes because it has not been set as inline
     expect(obj.getAttributes()).toEqual({
       id: 'test',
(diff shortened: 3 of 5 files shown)


src/common/index.ts
```diff
@@ -5,7 +5,7 @@ interface NOOP {}
 
 export type Debounced = Function & { cancel(): void };
 
-export type SetOptions = Backbone.ModelSetOptions & { avoidStore?: boolean };
+export type SetOptions = Backbone.ModelSetOptions & { avoidStore?: boolean; inline?: boolean };
```
artf, line 8: I think this one could be moved to `UpdateStyleOptions` type (I'll do it)

src/editor/config/config.ts
```diff
@@ -356,6 +357,11 @@ export interface EditorConfig {
    */
   commands?: CommandsConfig;
 
+  /**
+   * Configurations for keymaps
+   */
+  keymaps?: KeymapsConfig;
```
artf, line 363: 🙇‍♂️

test/specs/dom_components/model/Component.ts
```diff
@@ -291,13 +293,25 @@ describe('Component', () => {
...
+    obj.setStyle({ color: 'red' }); // Should be set on id
+    obj.setStyle({ display: 'flex' }, { inline: true }); // Should be set as inline
+
+    expect(obj.getStyle()).toEqual({
+      color: 'red',
+    });
+    expect(obj.getStyle({ inline: true })).toEqual({
+      display: 'flex',
+    });
+  });
```
artf, line 311: ❤️

artf (approved):
Makes all sense, I'd like to stay away from inline styles in general but this is actually a nice fix for the `inline` option.

Thank you also for the effort of making a good PR (context of the issue, comments in the right place, tests, etc.) 👏

Issue #6012, opened 2024-07-21
BUG: Input number value does not persist when using up/down arrows


aswinkumar863:
### GrapesJS version

- [X] I confirm to use the latest version of GrapesJS

### What browser are you using?

Chrome v126

### Reproducible demo link

https://codepen.io/aswinkumar863/pen/VwJaVYG

### Describe the bug

**How to reproduce the bug?**
1. Open the [demo](https://codepen.io/aswinkumar863/pen/VwJaVYG).
2. Click the "Click Me" section within the editor's body.
3. Navigate to the _Style Manager_ and select _General > Top_.
4. Enter any number into the input box without clicking outside it.
5. Click the up or down arrow.

**Note:** This behavior is observed only when the input field is not clicked outside of. Number traits are also affected by this issue.

**What is the expected behavior?**
The input number should increment or decrement from the currently entered value in the input box.

**What is the current behavior?**
Instead of stepping up/down from the input value, the input number is stepping up/down from the model value.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #6013, opened 2024-07-21
Fixed number input value does not persist when using up/down arrows, Issue #6012

aswinkumar863:
This PR addresses Issue #6012 where the number input value was not persisting when users used the up/down arrows to adjust the value.


artf (approved):
Good catch

PR #6018, opened 2024-07-25
Feature/data sources

danstarns:
Related: 

* https://github.com/GrapesJS/grapesjs/discussions/5956

This PR adds the ability to programmatically insert and inject DataRecords into components and their style.  

Documentation moved into source - see https://github.com/GrapesJS/grapesjs/blob/feature/data-sources/docs/modules/DataSources.md


artf:
Hey @lexoyo, since you've already worked on something similar, your insights into the API and the use of this module would be really helpful here.

lexoyo:
Sure! I'll take a look today
Thx for letting me know

lexoyo:
I really like the plucking part, and how you treat dynamic data as DOM nodes :clap: 

A few questions

1. I'm wondering how to have dynamic attributes and (css) classes?
1. As far as I understand, [Silex Lab's DataSource plugin](https://github.com/silexlabs/grapesjs-data-source) would use this new API for rendering the dynamic data only, is that right?

Also if you have time, I just created a discussion related to the integration of this API with the plugin: https://github.com/GrapesJS/grapesjs/discussions/6052

src/utils/mixins.ts
```diff
@@ -7,6 +7,35 @@ import { ObjectAny } from '../common';
...
+  return object.hasOwnProperty(value) ? [value] : stringToPath(value);
+}
+
+export const get = (object: ObjectAny, path: string | string[], def: any) => {
+  const paths = castPath(path, object);
+  const length = paths.length;
+  let index = 0;
+
+  while (object != null && index < length) {
+    object = object[`${paths[index++]}`];
```
lexoyo, line 34: I believe it is much faster to use .toString() (and easier to read)

lexoyo (commented):
This code is really clean and the use of backbone so natural :1st_place_medal:

artf:
Thanks for the comments @lexoyo the idea is to create the core business logic for binding generic data with the different parts of the editor (components, styles, traits, etc.). 

There won't be any UI around the module so I think your plugin is still relevant as you created already UI around the feature.

docs/modules/DataSources.md
```diff
@@ -0,0 +1,278 @@
+---
```
danstarns, line 1: See the description and docs here.

src/data_sources/model/DataRecord.ts
```diff
@@ -0,0 +1,164 @@
...
+
+      // @ts-ignore
+      super.set(attributeName, newValue, options);
+      return this;
+    } else {
+      // @ts-ignore
+      super.set(attributeName, value, options);
+      return this;
+    }
+  }
```
artf: I think this could be a bit less redundant
```js
set(...) {
  // ...
  let resultValue = value;
 
  if (onRecordSet && !options?.avoidTransformers) {
    resultValue = onRecordSet({...});
  }

  super.set(attributeName, resultValue, options);
  return this;
}
```

src/data_sources/model/DataSource.ts
```diff
@@ -0,0 +1,205 @@
...
+  removeRecord(id: string | number, opts?: RemoveOptions): DataRecord | undefined {
+    const onRecordDelete = this.transformers.onRecordDelete;
+    const record = this.getRecord(id);
+
+    if (record && onRecordDelete) {
+      onRecordDelete({ record });
+    }
+
+    return this.records.remove(id, opts);
+  }
```
artf, line 171: No more `setRecords` 🥲?

I think it will be quite common to reset the DataSource with new records

src/data_sources/model/DataSource.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  transformers?: DataSourceTransformers;
+}
+
+export interface DataSourceTransformers {
+  onRecordAdd?: (args: { record: DataRecordProps }) => DataRecordProps;
+  onRecordSet?: (args: { id: string | number; key: string; value: any }) => any;
+  onRecordDelete?: (args: { record: DataRecord }) => void;
+  onRecordRead?: (args: { record: DataRecord }) => DataRecord;
+}
```
artf: I'd put these inside `types.ts` to make the file a bit more readable, wdyt?

src/data_sources/model/DataVariable.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  defaults() {
+    return {
+      type: DataVariableType,
+      value: '',
+      path: '',
+    };
+  }
+
+  initialize(attrs: any, options: any) {
```
artf: any reason why you prefered `initialize` instead of the simple constructor?

test/specs/data_sources/model/TraitDataVariable.ts
```diff
@@ -0,0 +1,235 @@
...
+              type: DataVariableType,
+              value: 'default',
+              path: `${inputDataSource.id}.id1.value`,
+            },
+          },
+        ],
+      })[0];
+
+      const input = cmp.getEl();
+      expect(input?.getAttribute('value')).toBe('test-value');
```
artf, line 63: It's nice that we're testing the View/DOM but let's also test the value in the component Model (eg. `cmp.getAttributes().value`)

test/specs/data_sources/model/TraitDataVariable.ts
```diff
@@ -0,0 +1,235 @@
...
+      })[0];
+
+      const link = cmp.getEl() as HTMLLinkElement;
+      expect(link?.href).toBe('http://localhost/url-to-cat-image');
+
+      const testDs = dsm.get(inputDataSource.id);
+      testDs.getRecord('id1')?.set({ value: 'url-to-dog-image' });
+      expect(link?.href).toBe('http://localhost/url-to-dog-image');
+    });
+  });
```
artf, line 360: All the tests here cover the default trait behavior which is writing/reading to/from component attributes.

It would be also nice to test the [case of `changeProp`](https://grapesjs.com/docs/modules/Traits.html#add-traits-to-components), where the trait is writing/reading to/from component properties.

src/data_sources/index.ts
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * Remove data source.
+   * @param {String|[DataSource]} id Id of the data source.
+   * @returns {[DataSource]} Removed data source.
+   * @example
+   * const removed = dsm.remove('DS_ID');
+   */
+  remove(id: string | DataSource, opts?: RemoveOptions) {
+    return this.__remove(id, opts);
+  }
```
artf, line 122: I see that you removed from here `DataSources.getValue(path, defValue)` which was quite a useful API and way better than keeping the redundant one below in multiple places
```js
const [dsId, drId, keyPath] = stringToPath(path);
const ds = em?.DataSources.get(dsId);
const dr = ds?.records.get(drId);
const resValue = dr?.get(keyPath);
```
can we have it back, please? 🥲

Before it was also possible to get values directly from nested objects/arrays, eg. `{ id: 'record-id', someKey: { nested: [...] } }`, now it seems to be removed (quite crucial if you get data from something like a headless CMS).

src/trait_manager/model/Trait.ts
```diff
@@ -85,6 +102,32 @@ export default class Trait extends Model<TraitProperties> {
...
+    dataListeners.push({ obj: em, event: `${DataSourcesEvents.path}:${normPath}` });
+
+    dataListeners.forEach((ls) =>
+      this.listenTo(ls.obj, ls.event, () => {
+        const dr = dataVar.getDataValue();
+        this.updateValueFromDataVariable(dr);
+      }),
+    );
+    this.dataListeners = dataListeners;
+  }
```
artf: I see that we're using the same pattern in `StyleableModel.ts` and `ComponentDataVariableView.ts` (this one is actually more complete as here you don't update the value when for example a record is removed).

Maybe we should consolidate this logic in one single place? eg. `data_sources/model/DataVariable.ts` is the base class, `data_sources/model/TraitDataVariable.ts` might extend some logic if necessary

artf (changes requested):
Overall looks awesome 👏

Left a few comments to address regarding redundant parts

src/data_sources/model/DataSource.ts
danstarns: Good call, commited in https://github.com/GrapesJS/grapesjs/pull/6018/commits/5d6c479a99d18e9bf83d1e1a55249ed36fea7682

src/data_sources/model/DataVariable.ts
danstarns: No reason, good catch - removed.

test/specs/data_sources/model/TraitDataVariable.ts
danstarns: Good call, I made [changes to check the element and component model](https://github.com/GrapesJS/grapesjs/pull/6018/commits/fdd79e616b1c86938deeeb25eedfb2ef152873f1). During this, I found an error with the checkbox - see: 

* https://github.com/GrapesJS/grapesjs/pull/6095

test/specs/data_sources/model/TraitDataVariable.ts
```diff
@@ -0,0 +1,254 @@
...
+      expect(input?.getAttribute('checked')).toBe('true');
+
+      const testDs = dsm.get(inputDataSource.id);
+      testDs.getRecord('id1')?.set({ value: 'false' });
+
+      expect(input?.getAttribute('checked')).toBe('false');
+      // Not syncing - related to
+      // https://github.com/GrapesJS/grapesjs/discussions/5868
+      // https://github.com/GrapesJS/grapesjs/discussions/4415
+      // https://github.com/GrapesJS/grapesjs/pull/6095
```
danstarns, line 249: Related to @artf comment in this PR: 

* https://github.com/GrapesJS/grapesjs/pull/6018/files#r1730402702

src/data_sources/index.ts
danstarns: Good observation, I did not think about the nested object or arrays, added this back and some tests to assert nested object usage. 

* See commit https://github.com/GrapesJS/grapesjs/pull/6018/commits/da44545dcd47f10755d5bb20e15d202ea863ca6d

test/specs/data_sources/model/TraitDataVariable.ts
artf: I'm not sure but maybe with this test you found out a limitation in data sources with traits 😅 
After updating the data source record, I'd expect the model to be updated and the model update should trigger the view update accordingly.

src/data_sources/model/DataRecord.ts
```diff
@@ -0,0 +1,155 @@
...
+        : onRecordSet({
+            id: this.id,
+            key: attributeName as string,
+            value,
+          });
+
+    super.set(attributeName as string, newValue, options);
+
+    return this;
+  }
```
artf: You know, I like how you're managing onRecordSet here, makes sense to me. 

Quite curious why onRecordGet is not working the same way. You're actually suggesting mutating the model in onRecordGet, right? Are we sure this is the right way? 🤔

ps: Backbone is actually supporting also multiple properties updates, eg. `record.set({ key1: '', key2: '' })`, IMHO we have to support it here 🥲

danstarns: @artf please see commit: 

* https://github.com/GrapesJS/grapesjs/pull/6018/commits/5dc734bd6dbcc46a9ad18341eaf94093bab586a9

test/specs/data_sources/model/TraitDataVariable.ts
danstarns: @artf please see commit where I added data var usage and test with `changeProp`: 

* https://github.com/GrapesJS/grapesjs/pull/6018/commits/5dc734bd6dbcc46a9ad18341eaf94093bab586a9

src/data_sources/model/DataSource.ts
danstarns: [feat: add setRecords back](https://github.com/GrapesJS/grapesjs/pull/6018/commits/733656f80605cbabc517e80fea28cf5e34bcdda4)

src/trait_manager/model/Trait.ts
```diff
@@ -108,12 +108,21 @@ export default class Trait extends Model<TraitProperties> {
...
 
     prevListeners.forEach((ls) => this.stopListening(ls.obj, ls.event, this.updateValueFromDataVariable));
 
+    ds && dataListeners.push({ obj: ds.records, event: 'add remove reset' });
+    dr && dataListeners.push({ obj: dr, event: 'change' });
     dataListeners.push({ obj: dataVar, event: 'change:value' });
     dataListeners.push({ obj: em, event: `${DataSourcesEvents.path}:${normPath}` });
 
+    dataListeners.push(
+      { obj: dataVar, event: 'change:path change:value' },
```
artf:
```suggestion
      { obj: dataVar, event: 'change:path change:defaultValue' },
```

src/data_sources/model/DataVariableListenerManager.ts
artf, line 1: 🔝

src/dom_components/model/Component.ts
```diff
@@ -897,15 +906,24 @@ export default class Component extends StyleableModel<ComponentProperties> {
...
     const traits = this.traits;
     traits.each((trait) => {
-      if (!trait.changeProp) {
-        const name = trait.getName();
-        const value = trait.getInitValue();
+      const name = trait.getName();
+      const value = trait.getInitValue();
+
+      if (trait.changeProp) {
+        this.set(name, value);
```
artf, line 916: you know I'm not sure about this... mainly because with `changeProp` I'd expect the trait to get the value from the component. This probably would make sense when the property of the component is `undefined`.
wdyt?

src/dom_components/model/Component.ts
```diff
@@ -947,7 +965,7 @@ export default class Component extends StyleableModel<ComponentProperties> {
    * // append at specific index (eg. at the beginning)
    * someComponent.append(otherComponent, { at: 0 });
    */
-  append(components: ComponentAdd, opts: AddOptions = {}): Component[] {
+  append<T extends Component = Component>(components: ComponentAdd, opts: AddOptions = {}): T[] {
```
artf, line 968: 💜

src/dom_components/model/Component.ts
danstarns: Please see my reference in the test case

* https://github.com/GrapesJS/grapesjs/pull/6018/files/843492fbc7c6115b980d8e9dbcc47ffe8bb33b9a#r1737516791

test/specs/data_sources/model/TraitDataVariable.ts
```diff
@@ -0,0 +1,361 @@
...
+              type: DataVariableType,
+              defaultValue: 'default',
+              path: `${inputDataSource.id}.id1.value`,
+            },
+          },
+        ],
+      })[0];
+
+      let property = cmp.get('test-change-prop');
+      expect(property).toBe('I love grapes');
```
danstarns, line 352: Without said change how might this initial value look? Is it expected to have an initial value with `changeProp`? @artf

src/trait_manager/model/Trait.ts
danstarns: Listener logic is now inside the new class `DataVariableListenerManager`.

src/trait_manager/model/Trait.ts
danstarns: I am unsure about this one, what is `value` referencing here?

src/data_sources/model/DataRecord.ts
danstarns: @artf see commit: 

* https://github.com/GrapesJS/grapesjs/pull/6018/commits/d9e1bf53b21eac1da7ea2aac21e29a55538a42c2

I removed the following transformers: 

1. onRecordDelete
2. onRecordRead
3. onRecordSet

And now we only have one: `onRecordSetValue`. This is more than enough for someone to validate each entry and its called on record add and on record set, tested in `/tests/data_sources/transformers.ts`. I think this is a simpler API to start with.

src/trait_manager/model/Trait.ts
artf: This is backbone's default events on property changes. As you changed from `value` to `defaultValue`, also the event has to be changed

test/specs/data_sources/model/TraitDataVariable.ts
artf: mmm yeah, I think you're right... we should get the value from the trait if it's defined, but let's handle the case where the value of the trait is not defined.

```js
{
        tagName: 'div',
        type: 'default',
        'test-change-prop': 'initValue',
        traits: [
          {
            name: 'test-change-prop',
            type: 'text',
            changeProp: true,
          },
        ],
      }
```
makes sense?

src/data_sources/model/DataVariableListenerManager.ts
```diff
@@ -0,0 +1,62 @@
...
+    const normPath = stringToPath(path || '').join('.');
+    const prevListeners = this.dataListeners || [];
+    const [ds, dr] = this.em.DataSources.fromPath(path);
+
+    prevListeners.forEach((ls) => model.stopListening(ls.obj, ls.event, updateValueFromDataVariable));
+
+    const dataListeners: DataVariableListener[] = [];
+    ds && dataListeners.push({ obj: ds.records, event: 'add remove reset' });
+    dr && dataListeners.push({ obj: dr, event: 'change' });
+    dataListeners.push({ obj: dataVariable, event: 'change:value' });
```
artf, line 44: you have already this listener below `{ obj: dataVariable, event: 'change:path change:value' }`

src/data_sources/model/DataVariableListenerManager.ts
```diff
@@ -0,0 +1,62 @@
...
+    dataListeners.push({ obj: em, event: `${DataSourcesEvents.path}:${normPath}` });
```
artf, line 45: this is the same as the one below

src/data_sources/model/DataVariableListenerManager.ts
```diff
@@ -0,0 +1,62 @@
...
+    dataListeners.push(
+      { obj: dataVariable, event: 'change:path change:value' },
```
artf, line 47: the `value` is now `defaultValue`

src/data_sources/view/ComponentDataVariableView.ts
artf, line 1: perfect

test/specs/data_sources/model/ComponentDataVariable.ts
```diff
@@ -0,0 +1,298 @@
...
+    Pages.onLoad();
+    cmpRoot = Components.getWrapper()!;
+    const View = Components.getType('wrapper')!.view;
+    const wrapperEl = new View({
+      model: cmpRoot,
+      config: { ...cmpRoot.config, em },
+    });
+    wrapperEl.render();
+    fixtures = document.body.querySelector('#fixtures')!;
+    fixtures.appendChild(wrapperEl.el);
```
artf, line 30: nit: let's somehow consolidate these in `test/common.ts`?

test/specs/data_sources/model/ComponentDataVariable.ts
```diff
@@ -0,0 +1,298 @@
...
+      components: [
+        {
+          type: DataVariableType,
+          defaultValue: 'default',
+          path: 'ds1.id1.name',
+        },
+      ],
+    })[0];
+
+    expect(cmp.getEl()?.innerHTML).toContain('Name1');
```
artf, line 56: Let's also check the model of the component (eg. `cmp.getInnerHTML()`). 

The View part ensures it properly renders in the Canvas. The Model is more on the project JSON + HTML export side.

src/dom_components/model/Component.ts
artf: makes sense, let's just ensure the value is not undefined

artf (approved):
Minor comments but overall looks super clean and amazing 💜

src/data_sources/model/DataVariableListenerManager.ts
danstarns: * https://github.com/GrapesJS/grapesjs/pull/6098

src/data_sources/model/DataVariableListenerManager.ts
danstarns: * https://github.com/GrapesJS/grapesjs/pull/6098

src/data_sources/model/DataVariableListenerManager.ts
danstarns: * https://github.com/GrapesJS/grapesjs/pull/6098

test/specs/data_sources/model/ComponentDataVariable.ts
danstarns: * https://github.com/GrapesJS/grapesjs/pull/6099

src/trait_manager/model/Trait.ts
danstarns: * https://github.com/GrapesJS/grapesjs/pull/6098#issuecomment-2322320530

test/specs/data_sources/model/ComponentDataVariable.ts
danstarns: * https://github.com/GrapesJS/grapesjs/pull/6100

test/specs/data_sources/model/TraitDataVariable.ts
danstarns: * https://github.com/GrapesJS/grapesjs/pull/6101

src/dom_components/model/Component.ts
danstarns: * https://github.com/GrapesJS/grapesjs/pull/6101

PR #6041, opened 2024-08-06
refactor: remove dist from src

danstarns:
This PR removes the dist from source control. We should build the source from a unified machine and not expect it to come from a devs machine. 

@artf an open discussion on whether this would be a breaking change?  - Could users be pointing to the git URL?


artf (approved):
I think this was a mistake from the beginning keeping the `dist` folder in the source. Point directly to the GitHub URL file was also never recommended, so let's go with the cleanup 🥳

PR #6078, opened 2024-08-22
ci: add doc generation step into ci

danstarns:
This dependabot pr: 

* https://github.com/GrapesJS/grapesjs/pull/6071

It breaks the docs, but we did not catch it in CI because we don't build or compile the docs, lets add this doc generation runner into the quality checks so that any errors will bubble up and report feedback.


artf (approved):
👍

PR #6095, opened 2024-08-28
test: checkbox change propagation

danstarns:
Related: 

* https://github.com/GrapesJS/grapesjs/discussions/5868
* https://github.com/GrapesJS/grapesjs/discussions/4415
* https://github.com/GrapesJS/grapesjs/pull/6018#discussion_r1733726045

Adds a test to assert changes to the checkbox element get propagated to the component and model.

### DIFF
--- a/test/specs/issue_replications/checkbox-not-working.ts
+++ b/test/specs/issue_replications/checkbox-not-working.ts
@@ -1,0 +1,60 @@
+import Editor from '../../../src/editor/model/Editor';
+import ComponentWrapper from '../../../src/dom_components/model/ComponentWrapper';
+
+describe('Checkbox Behaviour', () => {
+  let em: Editor;
+  let fixtures: HTMLElement;
+  let cmpRoot: ComponentWrapper;
+
+  beforeEach(() => {
+    em = new Editor({
+      mediaCondition: 'max-width',
+      avoidInlineStyle: true,
+    });
+    document.body.innerHTML = '<div id="fixtures"></div>';
+    const { Pages, Components } = em;
+    Pages.onLoad();
+    cmpRoot = Components.getWrapper()!;
+    const View = Components.getType('wrapper')!.view;
+    const wrapperEl = new View({
+      model: cmpRoot,
+      config: { ...cmpRoot.config, em },
+    });
+    wrapperEl.render();
+    fixtures = document.body.querySelector('#fixtures')!;
+    fixtures.appendChild(wrapperEl.el);
+  });
+
+  afterEach(() => {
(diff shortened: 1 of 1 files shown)


test/specs/issue_replications/checkbox-not-working.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    const input = cmp.getEl() as HTMLInputElement;
+    expect(cmp.getAttributes().checked).toBe('true');
+    expect(input?.checked).toBe(true);
+    expect(input?.getAttribute('checked')).toBe('true');
+
+    input.checked = false;
+
+    expect(input?.checked).toBe(false);
+    expect(cmp.getAttributes().checked).toBe('false');
```
danstarns: Fails with: 

```
Summary of all failing tests
FAIL test/specs/issue_replications/checkbox-not-working.ts
  ● Checkbox Behaviour › init checkbox to true then change value and assert changes

    expect(received).toBe(expected) // Object.is equality

    Expected: "false"
    Received: "true"

      55 |
      56 |     expect(input?.checked).toBe(false);
    > 57 |     expect(cmp.getAttributes().checked).toBe('false');
         |                                         ^
      58 |     expect(input?.getAttribute('checked')).toBe('false');
      59 |   });
      60 | });

      at Object.toBe (test/specs/issue_replications/checkbox-not-working.ts:57:41)
```

Thoughts @artf ?

artf: So the View is reacting to the Model changes but not the other way.
In your test you're just updating the DOM, the Model is not aware of that.
This is how you would update the trait via Model:
```js
cmp.getTrait('checked').setValue(false);
```
this is expected to update the View/DOM

danstarns: Added your suggestion and the test still fails.

artf: The problem here is as you're using `valueTrue`/`valueFalse` as strings, and the DOM gets updated via `setAttribute('checked', value)` the DOM property `checked` doesn't actually change (as it accepts only a boolean).
So in your test case, only this can be valid: 
```js
expect(input?.getAttribute('checked')).toBe('false');
```

artf: IMHO there is no bug 
👋🎤

danstarns: Ok, I made a test to cover what we have, let's merge this as we are lacking coverage for checkboxes.

artf (approved):

PR #6099, opened 2024-08-30
refactor: reuse test setup for data sources

danstarns:
This PR cleans up the common test utils added in the Data Sources PR: 

* https://github.com/GrapesJS/grapesjs/pull/6018

Related Comment: 

* https://github.com/GrapesJS/grapesjs/pull/6018#discussion_r1738223036

### DIFF
--- a/test/common.ts
+++ b/test/common.ts
@@ -1,2 +1,26 @@
+import Editor from '../src/editor';
+
 // DocEl + Head + Wrapper
 export const DEFAULT_CMPS = 3;
+
+export function setupTestEditor() {
+  const editor = new Editor({
+    mediaCondition: 'max-width',
+    avoidInlineStyle: true,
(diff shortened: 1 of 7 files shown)


test/common.ts
```diff
@@ -1,2 +1,26 @@
+import Editor from '../src/editor';
+
 // DocEl + Head + Wrapper
 export const DEFAULT_CMPS = 3;
+
+export function setupTestEditor() {
```
danstarns, line 6: I think long term we should have the same function for each test we have, however for now as each test spawns a slightly different combination of editor and view etc, let's start with this and expand out.

artf: Yeah agree, in long term we can introduce more options to this helper

artf (approved):
🔝

PR #6127, opened 2024-09-05
refactor: remove extra build

danstarns:
We have an extra build step in the CI we don't need, and --force is not needed either.

### DIFF
--- a/.github/actions/setup-project/action.yml
+++ b/.github/actions/setup-project/action.yml
@@ -25,5 +25,5 @@ runs:
       run: pnpm install
       shell: bash
     - name: Build project
-      run: pnpm build --force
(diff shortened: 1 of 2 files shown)


artf (approved):
👍

One thing I'd recommend to check if we're still running `ts:build`, which generates TS files,  and `ts:check` which ensures the generated file is valid (sometimes it's happened the TS file had some issues)

danstarns:
> 👍
> 
> One thing I'd recommend to check if we're still running `ts:build`, which generates TS files, and `ts:check` which ensures the generated file is valid (sometimes it's happened the TS file had some issues)

https://github.com/GrapesJS/grapesjs/pull/6133

PR #6132, opened 2024-09-08
fix: remove id reassign in data sources

danstarns:
Related to: 

* https://github.com/GrapesJS/grapesjs/discussions/5956#discussioncomment-10559499

Removes the id reassign, adds coverage for discussed issue.


packages/core/test/specs/data_sources/__snapshots__/jsonplaceholder.ts.snap
```diff
@@ -0,0 +1,76 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`JsonPlaceholder Usage should render a list of comments from jsonplaceholder api 1`] = `
+"<body>
+  <div>
+    <h4>
+      <div>id labore ex et quam laborum</div>
+    </h4>
+    <p>
+    <div>1</div>
```
danstarns, line 10: @artf I wonder if we should allow people to change the dom component that gets injected, see here its always a `div` maybe we could allow in the data variable a way to specify the component to be injected like: 


```ts
editor.addComponents([
  {
    tagName: 'h1',
    type: 'text',
    components: [
      {
        type: 'data-variable',
        defaultValue: 'default',
        path: 'my-datasource.id1.content',
        domComponent: "div" <----- default
      },
    ],
  },
]);
```

Or maybe you don't want to inject a div inside a header so you change to a span. 


```ts
editor.addComponents([
  {
    tagName: 'h1',
    type: 'text',
    components: [
      {
        domComponent: "span"
      },
    ],
  },
]);
```

packages/core/test/specs/data_sources/jsonplaceholder.ts
```diff
@@ -0,0 +1,103 @@
+import DataSourceManager from '../../../src/data_sources';
+import ComponentWrapper from '../../../src/dom_components/model/ComponentWrapper';
+import { DataVariableType } from '../../../src/data_sources/model/DataVariable';
+import { DataSourceProps } from '../../../src/data_sources/types';
+import { setupTestEditor } from '../../common';
+import EditorModel from '../../../src/editor/model/Editor';
+import htmlFormat from 'pretty';
+
+async function getComments() {
+  const url = 'https://jsonplaceholder.typicode.com/posts/1/comments';
```
artf: mmm I'd avoid relying on external services in our tests, let's just get their data as example.

packages/core/test/specs/data_sources/__snapshots__/jsonplaceholder.ts.snap
artf: this should already work:
```js
{ 
type: 'data-variable', 
defaultValue: 'default', 
path: 'my-datasource.id1.content', 
tagName: "span"
},
```

artf:
@danstarns can you update the test as by suggestion?

danstarns:
@artf updated!

artf (approved):
🙇‍♂️

PR #6133, opened 2024-09-08
ci: add back ts:check

danstarns:
Related to: 

* https://github.com/GrapesJS/grapesjs/pull/6127

Adds tsbuild check back


artf (approved):
🙇‍♂️

Issue #4323, opened 2022-05-12
BUG: Performance issue while dragging component to canvas when you have 2k component on canvas


mohdSuhailCoditation:
### GrapesJS version

- [X] I confirm to use the latest version of GrapesJS

### What browser are you using?

Google Chrome Version 101.0.4951.41 (Official Build) (64-bit)

### Reproducible demo link

https://jsfiddle.net/Laze0pyg/

### Describe the bug

**How to reproduce the bug?**
1. Add 2k components using the below script.
2. `for(let i = 0;i<2000;i++)
    editor.addComponents('<div>ABC</div>');`
3. Try dragging Text Block on to the canvas.

**What is the expected behavior?**
There should be no lag.

**What is the current behavior?**
The green line placeholder is lagging

**JS Fiddle**
https://jsfiddle.net/Laze0pyg/

**Demo**

https://user-images.githubusercontent.com/100282281/168098195-9b69c333-df8b-43e0-bd64-4f7bee004b48.mp4

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

artf:
Yeah @GedMarc you're right but IMHO the Sorter could be definitely improved in such a case. Right now when the pointer goes over the component (eg. body) the Sorter calculates the dimensions and computed styles of all the children so I can see here a space for improvements.
Unfortunately, I don't have any ETA for such improvement, so if anybody wants to explore this path, PRs are welcome.

artf:
> But then when will it be enough? Slow down at 2k, then next someone report 3k, 5k, 10k, or more ...

Yeah I mean, the idea is if the HTML page itself is laggy (ml of DOM elements with styles and stuff) there is no point in improving as the editor preview should more or less show what you'd see outside of the editor.
But in the case of the demo, you can see how the lag starts only when you activate the component sorter, so that's why I see think there is room for improvements.

mohdSuhailCoditation:
Thanks @artf. I am happy to explore optimisation of sorter.js. It would be really helpful if you could provide some high level pointers and approach for the same.

lexoyo:
Hi that would be great to optimize this :+1:

I would suggest to use the browser's performance tool, but you probably know this

Thank you for looking into this @mohdSuhailCoditation

bgrand-ch:
@artf Does GrapesJS use Web Workers or virtual scrolling to optimise the rendering of a list of components or for large component (lot's of content with lot's of places taken? 🤔

[...]

PR #6149, opened 2024-09-16
Refactor sorter

mohamedsalem401:
For #6122 and possibly #4323 (as we will significantly reduce calls to the DOM).
### Changes
* Split the sorter into multiple classes.
* Change how we calculate drop position.
* Fix the issue with the StyleManager.

### Pre-Merge Checklist:
* **Canvas**
* **Style manager**
* **Layers**
* **Blocks**
* **Sorter event triggering:** Is the sorter triggering events as expected (e.g., `sorter:drag:start`)?
* Dropping image into the canvas
* D&D multiple components
* Rollback button
* Textable component D&D


packages/core/src/utils/sorter/types.ts
```diff
@@ -0,0 +1,107 @@
...
+ * Represents a collection of event handlers for sortable tree node events.
+ */
+export interface SorterEventHandlers<NodeType> {
+  onStartSort?: OnStartSortHandler<NodeType>;
+  onDragStart?: OnDragStartHandler;
+  onMouseMove?: OnMouseMoveHandler;
+  onDrop?: OnDropHandler<NodeType>;
+  onTargetChange?: OnTargetChangeHandler<NodeType>;
+  onPlaceholderPositionChange?: OnPlaceholderPositionChangeHandler;
+  onEndMove?: OnEndMoveHandler;
```
mohamedsalem401: Those might not be used currently, but I want them because they are callbacks that return good data from the sorter (e.g., all components being dragged), and other useful callbacks. Those can be deleted though.

mohamedsalem401:
@artf
- What do you think about the tests for the **old** sorter? Should I update them?
- There's a sorter in the **blocks view**, but it doesn't seem to be used anywhere that I know of. https://github.com/GrapesJS/grapesjs/blob/dev/packages/core/src/block_manager/view/BlocksView.ts

artf (approved):

PR #6155, opened 2024-09-19
Add ability to drag multiple selected components

mohamedsalem401:
When dragging multiple selected components on the canvas, only one component is currently moved. However, dragging from the drag icon in the toolbar correctly moves all selected components.

### DIFF
--- a/packages/core/src/dom_components/view/ComponentView.ts
+++ b/packages/core/src/dom_components/view/ComponentView.ts
@@ -164,7 +164,7 @@ TComp> {
     event.stopPropagation();
     event.preventDefault();
     this.em.Commands.run('tlb-move', {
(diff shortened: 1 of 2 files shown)


packages/core/src/dom_components/view/ComponentView.ts
```diff
@@ -164,7 +164,7 @@ TComp> {
     event.stopPropagation();
     event.preventDefault();
     this.em.Commands.run('tlb-move', {
-      target: this.model,
+      target: [...this.em.getEditor().getSelectedAll()],
```
artf: you can use `em.getSelectedAll()`

mohamedsalem401: @artf Done 👍

artf (approved):

PR #6160, opened 2024-09-21
feat: add DataSources to project json storage

danstarns:
This PR introduces the ability to store DataSources in the project JSON, allowing for persistent data across project saves and loads. Key additions include the `skipFromStorage` flag and updated store and load methods in the DataSourceManager.

### DIFF
--- a/docs/modules/DataSources.md
+++ b/docs/modules/DataSources.md
@@ -159,6 +159,59 @@ const testDataSource = {
 
 In this example, the `onRecordSetValue` transformer ensures that the `content` property is always an uppercase string.
 
+## Storing DataSources in Project JSON
+
+GrapesJS allows you to control whether a DataSource should be stored statically in the project JSON. This is useful for managing persistent data across project saves and loads.
+
+### Using the `shouldStoreInProject` Key
+
+When creating a DataSource, you can use the `shouldStoreInProject` key to specify whether it should be included in the project JSON.
+
+**Example: Creating a DataSource with `shouldStoreInProject`**
+
+```ts
+const persistentDataSource = {
+  id: 'persistent-datasource',
+  records: [
+    { id: 'id1', content: 'This data will be saved' },
+    { id: 'id2', color: 'blue' },
+  ],
+  shouldStoreInProject: true,
+};
+
+editor.DataSources.add(persistentDataSource);
+
+const temporaryDataSource = {
+  id: 'temporary-datasource',
+  records: [
+    { id: 'id1', content: 'This data will not be saved' },
+  ],
+  shouldStoreInProject: false, // This is the default if not specified
+};
+
+editor.DataSources.add(temporaryDataSource);
+```
+
+In this example, `persistentDataSource` will be included in the project JSON when the project is saved, while `temporaryDataSource` will not.
+
+### Benefits of Using `shouldStoreInProject`
+
+1. **Persistent Configuration**: Store configuration data that should persist across project saves and loads.
+2. **Default Data**: Include default data that should always be available in the project.
+3. **Selective Storage**: Choose which DataSources to include in the project JSON, optimizing storage and load times.
+
+### Accessing Stored DataSources
+
+When a project is loaded, GrapesJS will automatically restore the DataSources that were saved with `shouldStoreInProject: true`. You can then access and use these DataSources as usual.
+
+```ts
+// After loading a project
+const loadedDataSource = editor.DataSources.get('persistent-datasource');
+console.log(loadedDataSource.getRecord('id1').get('content')); // Outputs: "This data will be saved"
+```
+
+Remember that DataSources with `shouldStoreInProject: false` (or those without the key specified) will not be available after a project is loaded unless you add them programmatically.
+
 ## Benefits of Using DataSources
 
 DataSources are integrated with GrapesJS's runtime and BackboneJS models, enabling dynamic updates and synchronization between your data and UI components. This allows you to:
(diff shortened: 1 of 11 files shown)


docs/modules/DataSources.md
```diff
@@ -159,6 +159,59 @@ const testDataSource = {
...
+};
+
+editor.DataSources.add(persistentDataSource);
+
+const temporaryDataSource = {
+  id: 'temporary-datasource',
+  records: [
+    { id: 'id1', content: 'This data will not be saved' },
+  ],
+  shouldStoreInProject: false, // This is the default if not specified
```
artf: It's nice to have the ability to skip some dataSources from the storage but I see it as more of an exception. In most of the cases, dataSource has to be saved, so I'd put `true` as default.
ps. the property could also be renamed in `skipFromStorage`

packages/core/src/data_sources/index.ts
```diff
@@ -148,4 +149,59 @@ export default class DataSourceManager extends ItemManagerModule<ModuleConfig, D
...
+      Object.values(storedDataSources).forEach((ds) => {
+        this.add(ds, { silent: true });
+      });
+
+      return {
+        ...storedDataSources,
+        ...memoryDataSources,
+      };
+    }
+  }
```
artf: I'd expect this to be similar to other storable modules, eg.
```js
load(data: ProjectData) {
  return this.loadProjectData(data);
}
```
any reason of this custom implementation?

packages/core/src/data_sources/index.ts
```diff
@@ -148,4 +149,59 @@ export default class DataSourceManager extends ItemManagerModule<ModuleConfig, D
...
+        data[dataSource.id] = {
+          id: dataSource.id,
+          name: dataSource.get('name' as any),
+          records: dataSource.records.toJSON(),
+          shouldStoreInProject,
+        };
+      }
+    });
+
+    return { [this.storageKey]: data };
```
artf, line 170: It's a bit weird that we're storing data sources in an object and other storable modules are in an array. Any reason behind it?
Storing arbitrary strings in objects is also a common issue, eg. if the `id` is `constructor` it won't be saved properly 🥲

artf (changes requested):
Overall looks good but I'd suggest making the data source storable by default and I'd store them in an array instead of an object to make it more consistent and avoid some common issues with objects.

packages/core/src/data_sources/index.ts
danstarns: My thinking is that Data Sources could be fetched and defined before loading the project JSON.
Consider the example of loading a large JSON structure from an API. You may want to fetch and load the data into memory and into a data source first, then load the project JSON on top of it. Without this order of operations, you could experience flickering in the UI as it displays default values.

Additionally, you may have a combination of data sources stored in the project JSON, such as global styles, along with larger ones stored dynamically. This approach ensures that they are loaded correctly and not overridden depending on the order of operations.

packages/core/src/data_sources/index.ts
danstarns: I made it a key value simply so we can retrieve the data source by id faster than iterating over an array, it should not be too much of a performance problem so I am happy to change it to array like others.

packages/core/src/data_sources/index.ts
artf: Yeah, I guess there will be different logic and approaches to managing dynamic data sources but I'd leave that to the consumer (eg. hide the canvas until all the dynamic data sources are loaded).

All current storable modules should work in the same way. When you do `editor.loadProjectData(json)`, this is the request to load a completely new project, so you expect to `clear` all the current data sources and load those in the json (eg. if you have `dataSources: []`, that means the project doesn't contain any).
For sure someone will think one day to load partial data sources from another project but that logic should be handled on the consumer side (eg. the consumer loads the JSON of another project and decides what to add to the current one)

docs/modules/DataSources.md
danstarns: Changed to store by default and usage of `skipFromStorage` now. 

* https://github.com/GrapesJS/grapesjs/pull/6160/commits/d35f56f7fff766f2a508b502dde1c86ff28139f9

packages/core/src/data_sources/index.ts
danstarns: Changed here: 

* https://github.com/GrapesJS/grapesjs/pull/6160/commits/dbd660d8905b87e938c886bf7b5511823e6cab40

packages/core/src/data_sources/index.ts
danstarns: Sure, lets keep it standard, reverted here: 

* https://github.com/GrapesJS/grapesjs/pull/6160/commits/71f2cbfba7a1f004bf7a88842940d4caef7b066a

docs/modules/DataSources.md
```diff
@@ -159,6 +159,58 @@ const testDataSource = {
...
+editor.DataSources.add(temporaryDataSource);
+```
+
+In this example, `persistentDataSource` will be included in the project JSON when the project is saved, while `temporaryDataSource` will not.
+
+### Benefits of Using `skipFromStorage`
+
+1. **Persistent Configuration**: Store configuration data that should persist across project saves and loads.
+2. **Default Data**: Include default data that should always be available in the project.
+3. **Selective Storage**: Choose which DataSources to include in the project JSON, optimizing storage and load times.
```
artf, line 200: I guess this should be updated

artf (approved):
🚀

PR #6164, opened 2024-09-23
Prevent nesting instances of the same symbol

mohamedsalem401:
Currently, it's possible to nest an instance of a symbol within another instance of the same symbol. This can lead to unexpected behavior and potential issues.


artf (approved):
Really nice 👏
Thanks also for the additional tests on `canMove`

PR #6166, opened 2024-09-23
Fix CLI installation and add Windows compatibility for scripts

mohamedsalem401:
To use the GrapesJS framework via the CLI, we must first build the CLI itself. Therefore, we should add a build step for the CLI after the installation process.
```json
"start:js": "node node_modules/grapesjs-cli/dist/cli.js serve"
```

I also added cross-env before some script so that they will work on Windows machines.


danstarns:
Related: 

* https://github.com/GrapesJS/grapesjs/pull/6167

artf (approved):
Thanks @mohamedsalem401 🙇‍♂️
I removed for now `postinstall` as I think we want to avoid the consumers running build when they install grapesjs as a dependency.

We should probably highlight the CLI build step in the Development steps

PR #6167, opened 2024-09-23
feat: add windows to dev runners for DX

danstarns:
@mohamedsalem401 lets see if we can replicate the issues you are having locally in the CI. I cannot replicate without a windows machine.


artf:
@danstarns the build is now passing on windows but the CI fails on linting (`linebreaks to be 'LF' but found 'CRLF'  linebreak-style`) 🤔

artf:
@danstarns any update here?

PR #6173, opened 2024-09-25
Fix moving selected component after dragging an unselected component

mohamedsalem401:
#6155 implemented the ability to move multiple components after dragging. However, we should ensure that the dragged component is selected before attempting to move the selected components or else move the dragged component only.


artf (approved):
🚀

PR #6189, opened 2024-10-04
Feature detect document + fixes on configs

artf:
* Added the possibility to detect automatically if the passed HTML is a document
* Updated all configs in modules to avoid objects by reference

### DIFF
--- a/packages/core/src/panels/index.ts
+++ b/packages/core/src/panels/index.ts
@@ -27,7 +27,7 @@
  */
 import { Module } from '../abstract';
 import EditorModel from '../editor/model/Editor';
-import defaults, { PanelsConfig } from './config/config';
+import defConfig, { PanelsConfig } from './config/config';
 import Panel, { PanelProperties } from './model/Panel';
 import Panels from './model/Panels';
 import PanelsView from './view/PanelsView';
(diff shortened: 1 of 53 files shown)


packages/core/src/panels/index.ts
```diff
@@ -42,13 +42,8 @@ export default class PanelManager extends Module<PanelsConfig> {
...
    */
   constructor(em: EditorModel) {
-    super(em, 'Panels', defaults);
+    super(em, 'Panels', defConfig());
     this.panels = new Panels(this, this.config.defaults!);
-    for (var name in defaults) {
-      //@ts-ignore
-      if (!(name in this.config)) this.config[name] = defaults[name];
-    }
-    return this;
```
danstarns, line 51: My understanding of this change is to remove the need to do this loop. 

Making sure we always get a new copy of the default config.

artf: Correct, this was also an old way to copy configs, currently we're already doing that [here](https://github.com/GrapesJS/grapesjs/blob/922118f5acb29ac95faf7be8042119edf36cfd0e/packages/core/src/abstract/Module.ts#L63), so this one was pointless.

PR #6190, opened 2024-10-05
CssRule view datavar event propagation

danstarns:
This PR stores the view on a CSS rule and correctly attaches data variable listeners to update the model **and view** on change.

### DIFF
--- a/packages/core/src/css_composer/model/CssRule.ts
+++ b/packages/core/src/css_composer/model/CssRule.ts
@@ -94,7 +94,7 @@ export default class CssRule extends StyleableModel<CssRuleProperties> {
   config: CssRuleProperties;
   em?: EditorModel;
   opt: any;
-  view?: CssRuleView;
+  views: CssRuleView[] = [];
   dataVariableListeners: Record<string, DataVariableListenerManager> = {};
 
   defaults() {
@@ -121,7 +121,7 @@ export default class CssRule extends StyleableModel<CssRuleProperties> {
     this.em = opt.em;
     this.ensureSelectors(null, null, {});
     this.on('change', this.__onChange);
-    super.setStyle(this.get('style'));
+    this.setStyle(this.get('style'));
   }
 
   __onChange(m: CssRule, opts: any) {
@@ -364,13 +364,5 @@ export default class CssRule extends StyleableModel<CssRuleProperties> {
     }
 
     return true;
-  }
-
-  setView(cssRuleView: CssRuleView) {
-    super.setView(cssRuleView);
-  }
-
-  getView() {
-    return this.view;
   }
 }
--- a/packages/core/src/css_composer/model/CssRule.ts
+++ b/packages/core/src/css_composer/model/CssRule.ts
@@ -1,5 +1,5 @@
 import { isEmpty, forEach, isString, isArray } from 'underscore';
-import { Model, ObjectAny } from '../../common';
+import { Model, ObjectAny, View } from '../../common';
 import StyleableModel from '../../domain_abstract/model/StyleableModel';
 import Selectors from '../../selector_manager/model/Selectors';
 import { getMediaLength } from '../../code_manager/model/CssGenerator';
--- a/packages/core/src/css_composer/model/CssRule.ts
+++ b/packages/core/src/css_composer/model/CssRule.ts
@@ -6,6 +6,8 @@ import { getMediaLength } from '../../code_manager/model/CssGenerator';
 import { isEmptyObj, hasWin } from '../../utils/mixins';
 import Selector, { SelectorProps } from '../../selector_manager/model/Selector';
 import EditorModel from '../../editor/model/Editor';
+import CssRuleView from '../view/CssRuleView';
+import DataVariableListenerManager from '../../data_sources/model/DataVariableListenerManager';
 
 /** @private */
 export interface CssRuleProperties {
@@ -92,6 +94,8 @@ export default class CssRule extends StyleableModel<CssRuleProperties> {
   config: CssRuleProperties;
   em?: EditorModel;
   opt: any;
+  view?: CssRuleView;
+  dataVariableListeners: Record<string, DataVariableListenerManager> = {};
 
   defaults() {
     return {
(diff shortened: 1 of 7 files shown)


packages/core/test/specs/data_sources/model/StyleDataVariable.ts
```diff
@@ -183,7 +183,7 @@ describe('StyleDataVariable', () => {
         content: 'Hello World',
       })[0];
 
-      em.getEditor().CssComposer.addCollection([
+      const [rule] = em.getEditor().CssComposer.addCollection([
```
danstarns: @artf this pr is related to our discussion about getting CssRuleView to listen to a DataSource. 

Is it a 1-1 mapping from CssRule to CssRuleView? 

What assertions could we add in this file against the view? 

We already check pre and post update that the CSS is correct plus the rule.style is updated.

danstarns: I can compose a: 

```ts
const cssRuleView = new CssRuleView({ model: rule, config: rule.config });
```

From the rule, but it looks like there is no direct correlation to the view from the rule. 

In other DataSource test files we assert the component and the view like this for example: 

```ts
const style = cmp.getStyle();
expect(style).toHaveProperty('color', 'red'); // assuming this is the model
expect(cmp.getEl()?.innerHTML).toContain('Hello World'); // assuming this is the view
```

What other assertions could I make other than the: 

```ts
expect(rule.getStyle()).toHaveProperty('color', 'red');
expect(em.getEditor().getCss()).toContain(`${selector}{color:red;}`); // this is the view?
```

For a CSS Rule & its View?

danstarns: It looks here that the CssRuleView is created once, rendered, and then forgot about 

https://github.com/GrapesJS/grapesjs/blob/336f4b81e66206dcd3e067c8f94bf2d0e89533f3/packages/core/src/css_composer/view/CssRulesView.ts#L77-L84

I was wondering how we intended to attach watchers to these.

artf: Yeah, correct, we don't have any relation between CSSRule and CSSRuleView.
I think we can establish a similar relationship with Component and ComponentView (we're adding the view [here](https://github.com/GrapesJS/grapesjs/blob/025fd7dcb5dd47646ffe419b674f6e964dd1902f/packages/core/src/dom_components/view/ComponentView.ts#L81) and removing it [here](https://github.com/GrapesJS/grapesjs/blob/025fd7dcb5dd47646ffe419b674f6e964dd1902f/packages/core/src/dom_components/view/ComponentView.ts#L155)). 
It's a 1-to-many mapping (model could have multiple views) and once we have it we could probably trigger the dataSource update directly from the model side (eg. `model.views.forEach(view => view.renderStyles())`). wdyt?

packages/core/test/specs/data_sources/model/StyleDataVariable.ts
```diff
@@ -198,12 +200,19 @@ describe('StyleDataVariable', () => {
...
+      expect(rule.getStyle()).toHaveProperty('color', 'red');
       expect(em.getEditor().getCss()).toContain(`${selector}{color:red;}`);
+      expect(view?.el.innerHTML).toContain(`h1{color:red;}`);
 
       const ds = dsm.get(dsId);
       ds.getRecord(drId)?.set({ value: 'blue' });
 
+      expect(rule.getStyle()).toHaveProperty('color', 'blue');
       expect(em.getEditor().getCss()).toContain(`${selector}{color:blue;}`);
+      expect(view?.el.innerHTML).toContain(`h1{color:blue;}`);
```
danstarns, line 215: See now we have a `rule.getView()` method, that we assert updates to before and after data source update.

packages/core/src/css_composer/model/CssRule.ts
```diff
@@ -117,6 +121,7 @@ export default class CssRule extends StyleableModel<CssRuleProperties> {
     this.em = opt.em;
     this.ensureSelectors(null, null, {});
     this.on('change', this.__onChange);
+    super.setStyle(this.get('style'));
```
danstarns: Note** to call `.setStyle` on the styleable model to register the data var listeners. "The magic line". We could approach this better, open to thoughts....

packages/core/src/css_composer/view/CssRuleView.ts
```diff
@@ -13,7 +13,7 @@ export default class CssRuleView extends View<CssRule> {
     this.listenTo(model.get('selectors'), 'change', this.render);
   }
 
-  // @ts-ignore
+  /** @ts-ignore */
   tagName() {
     return 'style';
   }
```
danstarns: For some reason build fails if we don't add the comment: 

```
/** @ts-ignore */
```

I see we also have to do this in `ComponentView` file. 

Removing the ignore results in build error: 

```
Property 'tagName' in type 'ComponentView<TComp>' is not assignable to the same property in base type 'View<TComp, HTMLElement>'.
  Type '() => string' is not assignable to type 'string'.ts(2416)
Class 'View<TComp, HTMLElement>' defines instance member property 'tagName', but extended class 'ComponentView<TComp>' defines it as instance member function.
```


```
Property 'tagName' in type 'CssRuleView' is not assignable to the same property in base type 'View<CssRule, HTMLElement>'.
  Type '() => string' is not assignable to type 'string'.ts(2416)
Class 'View<CssRule, HTMLElement>' defines instance member property 'tagName', but extended class 'CssRuleView' defines it as instance member function.ts(2425)
```

Changing the method to a string then results in failing tests. So it looks like something somewhere is relying on `tagName` to be of type string not a method.

artf: yeah I don't remember if it was an issue on our side or the TS from the backbone, but it's ok, not a blocker

packages/core/src/css_composer/model/CssRule.ts
artf: I think for now it's fine

artf (approved):

PR #6200, opened 2024-10-08
Add dropzone

mohamedsalem401:
Use the following html for testing
```html
  <div id="gjs" style="position: relative; top: 32px; left: 32px; height: 0px; overflow: hidden">
    <style>
      /* CSS variables for width and height */
      :root {
        --inner-div-width: 32px;
        --inner-div-height: 32px;
      }

      /* Outer container */
      .outer-div {
        display: flex;
        flex-direction: column;
        border: 1px solid black;
        /* Optional border for outer div */
        width: fit-content;
      }

      /* Row styling */
      .row {
        display: flex;
        flex-direction: row;
      }

      /* Inner divs in each row */
      .inner-div {
        width: var(--inner-div-width);
        height: var(--inner-div-height);
        border: 2px solid red;
      }
    </style>
    <div class="outer-div">
      <div class="row">
        <div class="inner-div"></div>
        <div class="inner-div"></div>
        <div class="inner-div"></div>
      </div>
      <div class="row">
        <div class="inner-div"></div>
        <div class="inner-div"></div>
        <div class="inner-div"></div>
      </div>
    </div>
  </div>
```

### DIFF
--- a/packages/core/src/canvas/index.ts
+++ b/packages/core/src/canvas/index.ts
@@ -499,7 +499,7 @@ export default class CanvasModule extends Module<CanvasConfig> {
    * @return {Object}
    * @private
    */
-  getMouseRelativeCanvas(ev: MouseEvent, opts: any) {
+  getMouseRelativeCanvas(ev: MouseEvent | { clientX: number; clientY: number }, opts: any) {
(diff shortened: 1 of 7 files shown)


artf (approved):
💜

PR #6210, opened 2024-10-11
feat: add telemetry tracking

danstarns:
You can disable this with options.telemetry: false. 

We aim on adding metrics to the lib to provide a better service.

### DIFF
--- a/docs/guides/Telemetry.md
+++ b/docs/guides/Telemetry.md
@@ -27,7 +27,8 @@ We use data to:
 You can opt-out of data collection by setting the `telemetry` option to `false` when initializing GrapesJS:
 
 ```js
-const editor = new Editor({
+const editor = grapesjs.init({
+  // ...
   telemetry: false,
 });
 ```
--- a/docs/guides/Telemetry.md
+++ b/docs/guides/Telemetry.md
@@ -1,0 +1,33 @@
+---
+title: GrapesJS Telemetry
+---
+
+# GrapesJS Telemetry
+
+We collect and use data to improve GrapesJS. This page explains what data we collect and how we use it.
+
+## What data we collect
+
+We collect the following data:
+
+- **domain**: The domain of the website where GrapesJS is used.
+- **version**: The version of GrapesJS used.
+- **timestamp**: The time when the editor is loaded.
+
+## How we use data
+
+We use data to:
+
+- **Improve GrapesJS**: We use data to improve GrapesJS. For example, we use data to identify bugs and fix them.
+- **Analyze usage**: We use data to analyze how GrapesJS is used. For example, we use data to understand which features are used most often.
+- **Provide support**: We use data to provide support to users. For example, we use data to understand how users interact with GrapesJS.
+
+## How to opt-out
+
+You can opt-out of data collection by setting the `telemetry` option to `false` when initializing GrapesJS:
+
+```js
+const editor = new Editor({
+  telemetry: false,
+});
+```
--- a/packages/core/src/editor/config/config.ts
+++ b/packages/core/src/editor/config/config.ts
@@ -428,6 +428,13 @@ export interface EditorConfig {
    */
   colorPicker?: ColorPickerOptions;
   pStylePrefix?: string;
+
+  /**
+   * Telemetry options
+   * Default: true
+   */
+
+  telemetry?: boolean;
 }
 
 export type EditorConfigKeys = keyof EditorConfig;
@@ -506,6 +513,7 @@ const config: () => EditorConfig = () => ({
   textViewCode: 'Code',
   keepUnusedStyles: false,
   customUI: false,
+  telemetry: true,
 });
 
 export default config;
--- a/packages/core/src/editor/config/config.ts
+++ b/packages/core/src/editor/config/config.ts
@@ -433,7 +433,6 @@ export interface EditorConfig {
    * Telemetry options
    * Default: true
    */
-
   telemetry?: boolean;
 }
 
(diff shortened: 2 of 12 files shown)


packages/core/src/editor/model/Editor.ts
```diff
@@ -1136,4 +1137,60 @@ export default class EditorModel extends Model {
...
+    });
+  }
+
+  trackEditorLoad() {
+    this.on('load', () => {
+      if (this._config.telemetry) {
+        this.sendTelemetryData();
+      }
+    });
+  }
```
danstarns: @artf what is the best way to hook into editor load, should we do it on the main editor instance or the model, and what event is appropriate for the "EDITOR:LOAD" event tracking?

packages/core/test/setup.js
```diff
@@ -16,6 +16,7 @@ const localStorage = {
 
 global._ = _;
 global.__GJS_VERSION__ = '';
+global.__STUDIO_URL__ = '';
```
danstarns: Re - inject a static api key here 👀

packages/core/test/specs/editor/telemetry.ts
```diff
@@ -0,0 +1,126 @@
...
+  const triggerLoadAndWait = (editor: Editor) => {
+    return new Promise<void>((resolve) => {
+      editor.on('telemetry:sent', () => {
+        resolve();
+      });
+      editor.getModel().trigger('load');
+    });
+  };
+
+  test('Telemetry is sent when enabled', async () => {
```
danstarns, line 79: Only this test in the file is working, the rest hang around - something to do with timers, hoping to shed light on what events to listen to from this comment: 

* https://github.com/GrapesJS/grapesjs/pull/6210/files#r1796327194

packages/core/src/editor/config/config.ts
```diff
@@ -428,6 +428,13 @@ export interface EditorConfig {
    */
   colorPicker?: ColorPickerOptions;
   pStylePrefix?: string;
+
+  /**
+   * Telemetry options
+   * Default: true
+   */
+
+  telemetry?: boolean;
```
danstarns:
```suggestion
   */
  telemetry?: boolean;
```

packages/core/test/specs/editor/telemetry.ts
artf: Check this one https://github.com/GrapesJS/grapesjs/pull/6195

packages/core/src/editor/model/Editor.ts
```diff
@@ -1136,4 +1137,69 @@ export default class EditorModel extends Model {
...
+
+    if (!enableDevTelemetry && (hostName === 'localhost' || hostName.includes('localhost'))) {
+      // Don't send telemetry data for localhost
+      return;
+    }
+
+    const sessionKeyPrefix = 'gjs_telemetry_sent_';
+
+    // @ts-ignore
+    const version = __GJS_VERSION__;
```
artf: no need for this one, you can recover the version from `this.config.grapesjs.version`

packages/core/src/editor/model/Editor.ts
```diff
@@ -1136,4 +1137,69 @@ export default class EditorModel extends Model {
...
+    const sessionKey = `${sessionKeyPrefix}${version}`;
+
+    if (sessionStorage.getItem(sessionKey)) {
+      console.log(`Telemetry already sent for version ${version} this session`);
```
artf: cleanup the log

packages/core/src/editor/model/Editor.ts
artf: Let's trigger `sendTelemetryData` from [here](https://github.com/GrapesJS/grapesjs/blob/d74c3f18a873bac5bd9658659932acf165c91ef5/packages/core/src/editor/view/EditorView.ts#L10-L17). `EDITOR:LOAD` sounds good to me

packages/core/src/editor/model/Editor.ts
```diff
@@ -1136,4 +1137,69 @@ export default class EditorModel extends Model {
...
+      return;
+    }
+
+    // @ts-ignore
+    const url = __STUDIO_URL__;
+    const path = '/api/gjs/telemetry/collect';
+
+    (async () => {
+      const response = await fetch(`${url}${path}`, {
```
artf: let's make `sendTelemetryData` async so you can avoid IIFE here?

docs/guides/Telemetry.md
```diff
@@ -0,0 +1,33 @@
...
+- **Provide support**: We use data to provide support to users. For example, we use data to understand how users interact with GrapesJS.
+
+## How to opt-out
+
+You can opt-out of data collection by setting the `telemetry` option to `false` when initializing GrapesJS:
+
+```js
+const editor = new Editor({
+  telemetry: false,
+});
```
artf:
```js
  // ...
  telemetry: false,
});
```

packages/core/src/editor/model/Editor.ts
danstarns: I cant seem to get this to work, or find a reference for it in the code. 

```
TypeError: Cannot read properties of undefined (reading 'version')

      1153 |
      1154 |     // @ts-ignore
    > 1155 |     const version = this.config.grapesjs.version;
           |                                          ^
      1156 |     const sessionKey = `${sessionKeyPrefix}${version}`;
      1157 |
      1158 |     if (sessionStorage.getItem(sessionKey)) {
```

packages/core/src/editor/model/Editor.ts
danstarns: The IIFE is so we don't block sending telemetry data, the function will be added to the queue and resolved when appropriate. Also its an IIFE because I am not sure if backbone event listeners can be async.

packages/core/test/specs/editor/telemetry.ts
danstarns: Thanks, moving to editor view load and using your test utils works.

artf (approved):
💜

PR #6219, opened 2024-10-15
A few fixes/improvements for the DataSource module

artf:


packages/core/src/data_sources/index.ts
```diff
@@ -177,4 +177,9 @@ export default class DataSourceManager extends ItemManagerModule<ModuleConfig, D
   load(data: any) {
     return this.loadProjectData(data);
   }
+
+  postLoad() {
+    const { em, all } = this;
+    em.listenTo(all, collectionEvents, (m, c, o) => em.changesUp(o || c));
+  }
```
artf, line 184: trigger editor update on data source collection changes

packages/core/src/data_sources/model/DataSource.ts
```diff
@@ -74,6 +74,7 @@ export default class DataSource extends Model<DataSourceProps> {
     }
 
     this.listenTo(this.records, 'add', this.onAdd);
+    this.listenTo(this.records, collectionEvents, this.handleChanges);
```
artf, line 77: Trigger editor updates on data record collection changes.

I think some data sources (eg. with data loaded asynchronously) could be updated without triggering any update.

packages/core/src/domain_abstract/model/StyleableModel.ts
```diff
@@ -151,22 +151,11 @@ export default class StyleableModel<T extends ObjectHash = any> extends Model<T>
         model: this,
         em: this.em!,
         dataVariable: dataVar,
-        updateValueFromDataVariable: (newValue: string) => this.updateStyleProp(styleProp, newValue),
+        updateValueFromDataVariable: () => this.updateView(),
```
artf, line 154: no need to update the model as it's already resolved inside `getStyle`

PR #6229, opened 2024-10-20
Refactor data variable listeners to a generic dynamic variable

mohamedsalem401:
Refactor datasource variables before adding conditional data variables


packages/core/src/domain_abstract/model/StyleableModel.ts
```diff
@@ -114,9 +114,17 @@ export default class StyleableModel<T extends ObjectHash = any> extends Model<T>
...
-        this.manageDataVariableListener(styleDataVariable, key);
+        const dynamicType = styleValue.type;
+        let styleDynamicVariable;
+        switch (dynamicType) {
+          case DataVariableType:
+            styleDynamicVariable = new StyleDataVariable(styleValue, { em: this.em });
+            break;
+          default:
+            throw new Error(`Invalid data variable type. Expected '${DataVariableType}', but found '${dynamicType}'.`);
+        }
```
mohamedsalem401, line 125: I know that I'm adding a whole switch statment for just a single check but I'll later add conditional type.

artf (approved):
Nice 👍

PR #6238, opened 2024-10-21
package: pin webpack version for vuepress

danstarns:
Vuepress and dependants @ 1.9.10 depend on Webpack 4. https://www.npmjs.com/package/@vuepress/core?activeTab=code see package json.

We can upgrade to v2 of vuepress, however, [they are still on rc flags. ](https://github.com/vuepress/core/blob/main/CHANGELOG.md)

This PR fixes us to webpack v4 in the docs directory and removes the build filter in the root package JSON.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -28,6 +28,7 @@
     "postcss": "8",
     "sass": "1.42.1",
     "vuepress": "1.9.10",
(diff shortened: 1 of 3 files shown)


artf (approved):
💜

PR #6239, opened 2024-10-21
Add data condition variable

mohamedsalem401:
- Add condition evaluation
- Add logical groups
- Add nested conditions
- Add support for some comparison statments ( string comparison - numbers comparison - ... )


artf (approved):
🚀

muarifer:
Hi. Could you give me an example usage?

artf:
@muarifer we'll work on docs later, for now, I'd suggest you check the provided test use cases to see what is possible to do

PR #6270, opened 2024-10-25
Add datasource support to conditional variables

mohamedsalem401:
- Add data variables support to conditional variables
- Listen to changes from data variables


artf (approved):
🚀

PR #6276, opened 2024-10-29
generate html should keep id when script-export is set

Singwai:
Question:
With cleanId enabled, should we check both script and script-export similar to match the js code generator logic in the html custom attributes function? 

[Here](https://github.com/Singwai/grapesjs/blob/2055896922039ae1bc7663ac15582cf82d663c5c/packages/core/src/code_manager/model/JsGenerator.ts#L21) is the reference to ths jsGenerator. I think the getHtml code with cleanId might be have a chance to lost the id during export depending on the component's configuration.

Potential STR:
 A component with script-export defined and script is empty
Without explicitly setting the id in the component attribute
Call  `editor.getHtml({cleanId: true})`


artf (approved):

artf:
Makes sense @Singwai thank you 🙇‍♂️

Issue #6271, opened 2024-10-25
BUG: Drag and drop of components doesn't work on latest Firefox


padcom:
### GrapesJS version

- [X] I confirm to use the latest version of GrapesJS

### What browser are you using?

Firefox 131.0 (64-bit) on Linux and Mac

### Reproducible demo link

https://grapesjs.com/demo

### Describe the bug

**How to reproduce the bug?**
1. Start dragging the "Hover me" button
2. Try to put it above the "All text blocks could be edited easily..."

**What is the expected behavior?**
- The button can be dropped anywhere
- The layer manager can be used to reorganize elements

**What is the current behavior?**
- The button can be only dropped at the beginning of the container
- Dragging within the layer manager has the same limitation

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

padcom:
I've tracked down the breaking change in Firefox to version 109.0. It's the first one that has the problem. Version 108.0.2 works fine. Hope that helps.

padcom:
I have also tested 133.0a1 nightly from 2024-10-24-21-16-15-mozilla-central and the problem is also there.

padcom:
When can we expect this fix to be released?

artf:
@padcom https://github.com/GrapesJS/grapesjs/releases/tag/v0.22.2

padcom:
You guys are awesome! And not a moment too soon :) I have a presentation Today in the afternoon and now I'll be able to show it working on Firefox too!

PR #6278, opened 2024-10-29
Fix drag and drop issue on firefox

mohamedsalem401:
For https://github.com/GrapesJS/grapesjs/issues/6271
If the `HTMLElement` is created or passed from an iframe, it will not be recognized as the same HTMLElement type.


artf (approved):
🙇‍♂️ thanks @mohamedsalem401

PR #6282, opened 2024-10-30
Add conditional variables

mohamedsalem401:
## **Features Implemented**  
- **Add Traits Conditional Variables**  
- **Add Components Conditional Variables**  
- **Add Styles Conditional Variables**

---

## **Notes**  
1. **Handling Conditional Components**  
   - The return value for a conditional component will be **added as a child** of the conditional component.  

   **Example:**  
   ```ts
   const component = cmpRoot.append({
     type: ConditionalVariableType,
     condition: true,
     ifTrue: '<div>This will be a child of the conditional component</div>',
   })[0];
   ```

2. **Saving Changes to Conditional Component Children**  
   - Any changes made to the child of a conditional component will be **persisted to storage**.  

   **Example (with updated styles):**  
   ```ts
   {
     type: ConditionalVariableType,
     condition: true,
     ifTrue: {
       type: 'div',
       content: 'This will be a child of the conditional component',
       style: {
         color: 'red',
       },
     },
   }
   ```

---

### **Explanation**  
- In the example above, if you **add or modify styles** (e.g., setting `color: red`) on the child `<div>`, the changes will be **saved to storage** and reflected whenever the component is reloaded or rendered again.


packages/core/src/domain_abstract/model/StyleableModel.ts
```diff
@@ -150,10 +143,35 @@ export default class StyleableModel<T extends ObjectHash = any> extends Model<T>
     return newStyle;
   }
 
+  private isDynamicValue(styleValue: any) {
+    return typeof styleValue === 'object' && [DataVariableType, ConditionalVariableType].includes(styleValue.type);
+  }
+
+  private resolveDynamicValue(styleValue: any) {
```
artf: replace `any`

packages/core/src/dom_components/model/Component.ts
```diff
@@ -769,10 +771,31 @@ export default class Component extends StyleableModel<ComponentProperties> {
...
+        let dataVariable: TraitDataVariable | DataCondition;
+
+        switch (true) {
+          case value instanceof DataCondition:
+          case value instanceof TraitDataVariable:
+            dataVariable = value;
+            break;
+
+          case (value as any).type === ConditionalVariableType: {
+            const { condition, ifTrue, ifFalse } = value as any;
```
artf: I guess here the `any` could be avoided

packages/core/src/trait_manager/model/Trait.ts
```diff
@@ -63,8 +64,15 @@ export default class Trait extends Model<TraitProperties> {
...
+        case ConditionalVariableType: {
+          const { condition, ifTrue, ifFalse } = this.attributes.value;
+          this.dynamicVariable = new DataCondition(condition, ifTrue, ifFalse, { em: this.em });
+          break;
+        }
         default:
-          throw new Error(`Invalid data variable type. Expected '${DataVariableType}', but found '${dataType}'.`);
+          throw new Error(
+            `Invalid data variable type. Expected '${DataVariableType} or ${ConditionalVariableType}', but found '${dataType}'.`,
+          );
```
artf: You know, I think we should avoid triggering errors here. The consumer could use any kind of custom value he wants (also objects).

packages/core/src/dom_components/model/Component.ts
```diff
@@ -769,10 +771,31 @@ export default class Component extends StyleableModel<ComponentProperties> {
       }
     }
 
-    const attrDataVariable = this.get('attributes-data-variable');
+    const attrDataVariable = this.get('attributes-dynamic-value');
```
artf: let's put `attributes-dynamic-value` in a const variable?! 
I think it will be easier to track its usage.

packages/core/test/specs/data_sources/model/conditional_variables/ComponentConditionalVariable.ts
```diff
@@ -0,0 +1,286 @@
...
+      },
+    })[0];
+    expect(component).toBeDefined();
+    expect(component.get('type')).toBe(ConditionalVariableType);
+    expect(component.getInnerHTML()).toBe('<h1>some text</h1>');
+
+    const childComponent = component.components().at(0);
+    expect(childComponent).toBeDefined();
+    expect(childComponent.get('type')).toBe('text');
+    expect(childComponent.getInnerHTML()).toBe('some text');
```
artf: Here you're checking only the "model" of the component, let's place also a check for the "view" so we know it's rendered as expected in the canvas.

Same for the test cases below

packages/core/test/specs/data_sources/model/conditional_variables/ComponentConditionalVariable.ts
```diff
@@ -0,0 +1,286 @@
...
+    })[0];
+    dsm.get('ds1').getRecord('left_id')?.set('left', 'Diffirent value');
+
+    const childComponent = component.components().at(0);
+    expect(childComponent).toBeDefined();
+    expect(childComponent.get('type')).toBe('text');
+    expect(childComponent.getInnerHTML()).toBe('False value');
+  });
+
+  it('should change with changing value of data-source', () => {
```
artf: this seems to be the same test above

packages/core/test/specs/data_sources/model/conditional_variables/ConditionalTraits.ts
```diff
@@ -0,0 +1,174 @@
...
+              right: -1,
+            },
+            ifTrue: 'Some title',
+          },
+        },
+      ],
+    })[0];
+
+    expect(component).toBeDefined();
+    expect(component.getTrait('title').get('value')).toBe('Some title');
```
artf: let's also check the component attribute model (`cmp.getAttributes().title`) and its rendered view.

Same for the test cases below

packages/core/test/specs/data_sources/model/conditional_variables/ConditionalTraits.ts
```diff
@@ -0,0 +1,174 @@
...
+            value: {
+              type: ConditionalVariableType,
+            },
+          },
+        ],
+      });
+    }).toThrow(MissingConditionError);
+  });
+
+  it('should store traits with conditional values correctly', () => {
```
artf, line 139: let's also check everything works as expected when we load some projectData

packages/core/test/specs/data_sources/model/conditional_variables/ConditionalTraits.ts
```diff
@@ -0,0 +1,174 @@
...
+      ifTrue: 'Positive',
+    };
+    cmpRoot.append({
+      tagName: 'h1',
+      type: 'text',
+      traits: [
+        {
+          type: 'text',
+          name: 'dynamicTrait',
+          value: conditionalTrait,
```
artf, line 156: by default trait changes are added to the component as attributes, let's also ensure conditions are working with traits and `changeProp: true` (in that case trait changes are added to the component properties)

artf (commented):
overall looks really good 👏

packages/core/src/trait_manager/model/Trait.ts
mohamedsalem401: Done and I've also added a test for custom and `object` values.

packages/core/test/specs/data_sources/model/conditional_variables/ConditionalTraits.ts
mohamedsalem401: I've added some tests for that. However, there's a current limitation: the implementation modifies both the attribute and property values, which is unintended behavior. This issue also affects data source variables. I'll fix this in a new PR.

artf (approved):
🚀

PR #6283, opened 2024-10-31
Fix layer manager drag and drop

### LINKED ISSUE #6281: BUG: Layer Manager does not allow reordering on same level. only nesting is allowed on top level(right under body) older versions (e.g: 0.21.8
### GrapesJS version

latest chrome, firefox, edge

### Reproducible demo link

https://grapesjs.com/demo

### Describe the bug

**How to reproduce the bug?**
1. drag a acomponent in layer manager and try to put it between other items under body
2. not allowed. nesting is allowed

... both nesting and reordering shall be enabled

...only nesting enabled on top level

### Code of Conduct

mohamedsalem401:
For https://github.com/GrapesJS/grapesjs/issues/6281

We need more area to allow dragging in-between components in the layer manager
image

### DIFF
--- a/packages/core/src/utils/sorter/LayersComponentNode.ts
+++ b/packages/core/src/utils/sorter/LayersComponentNode.ts
@@ -1,6 +1,11 @@
 import { BaseComponentNode } from './BaseComponentNode';
 
 export default class LayersComponentNode extends BaseComponentNode {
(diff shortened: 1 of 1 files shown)


artf (approved):
🙇‍♂️

PR #6312, opened 2024-11-18
fix: ensure the pointer-init class is removed when updating the "locked" state

WilliamDASILVA:
The locking mechanism does not work as expected when unlocking an element.  
This is due to the fact that both `pointer-init` and `no-pointer` classes are present on the element.  
When unlocking and then locking the element, we expect the element to remain locked.  

This PR ensures that `pointer-init` is removed along with the other classes to avoid conflict.
 
 
 Reproduction in https://app.grapesjs.com/
2024-11-18 at 10 49 22 - Tan Lark


artf (approved):
Good catch @WilliamDASILVA thank you 🙇‍♂️

PR #6316, opened 2024-11-19
test: add assert for component:add event

danstarns:
This adds a test to ensure only 1 event is fired on component addition. 

Related: 

* https://github.com/GrapesJS/grapesjs/discussions/6314


artf:
Closing this as no reproducible steps for the issue were provided

PR #6351, opened 2024-12-10
Add Components dynamic values ( props - attributes )

mohamedsalem401:
- Disable dynamic values for traits ( until we decide if we should keep it ) ( Update: we won't )
- Add dynamic values for props and attributes
- Add unit tests for both dynamic values for props and attributes
- Make traits dependent on component values for props and attributes


packages/core/test/specs/data_sources/dynamic_values/attributes.ts
```diff
@@ -0,0 +1,247 @@
...
+      changeDataSourceValue(dsm);
+      testAttribute(cmp, 'dynamicAttribute', 'changed-value');
+    });
+
+    test('(Component.setAttributes) dynamic attributes should listen to the latest dynamic value', () => {
+      const dataSource = {
+        id: 'ds_id',
+        records: [
+          { id: 'id1', value: 'test-value' },
+          { id: 'id2', value: 'test-value' },
```
artf: Let's use  something like `test-value2` for `id2`

packages/core/test/specs/data_sources/model/conditional_variables/ConditionalTraits.ts
```diff
@@ -165,12 +165,13 @@ describe('TraitConditionalVariable', () => {
...
-      dynamicTrait: conditionalTrait,
-    });
+    // expect(storedComponent[dynamicAttrKey]).toEqual({
+    //   dynamicTrait: conditionalTrait,
+    // });
   });
 
-  it('should load traits with conditional values correctly', () => {
+  // TODO: Should we keep or delete saving triats with dynamic values?
+  it.skip('should load traits with conditional values correctly', () => {
```
artf: I'd say it's important to preserve dynamic definitions, otherwise, you will lose data source connections on save/load, right?

packages/core/src/dom_components/model/Component.ts
```diff
@@ -662,10 +695,15 @@ export default class Component extends StyleableModel<ComponentProperties> {
    * component.addAttributes({ 'data-key': 'value' });
    */
   addAttributes(attrs: ObjectAny, opts: SetAttrOptions = {}) {
+    const areStaticAttributes = DynamicValueWatcher.areStaticValues(attrs);
+    this.componentDVListener.removeAttributes(Object.keys(attrs));
+    const evaluatedAttributes = areStaticAttributes ? attrs : DynamicValueWatcher.getStaticValues(attrs, this.em);
+    this.componentDVListener.watchAttributes(attrs);
+
```
artf: Honestly, I'd expect all the logic to be managed inside `setAttributes` (eg. clearing removed dynamic attributes, listening new dynamic attributes)

packages/core/src/dom_components/model/Component.ts
```diff
@@ -331,6 +337,36 @@ export default class Component extends StyleableModel<ComponentProperties> {
...
+    const areStaticAttributes = DynamicValueWatcher.areStaticValues(attributes);
+    const evaluatedAttributes = areStaticAttributes
+      ? attributes
+      : ComponentDynamicValueListener.evaluateComponentDef(attributes, this.em);
+
+    if (!options.skipWatcherUpdates) {
+      this.componentDVListener?.watchProps(attributes);
+    }
+
+    return super.set(evaluatedAttributes, options);
```
artf, line 367: I think you're doing the check for attributes but not the options here. If you do `cmp.set({...}, options)`  the options won't be used.

packages/core/src/dom_components/model/ComponentDynamicValueListener.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  private createPropertyUpdater() {
+    return (key: string, value: any) => {
+      this.component.set(key, value, { skipWatcherUpdates: false });
+    };
+  }
+
+  private createAttributeUpdater() {
+    return (key: string, value: any) => {
+      this.component.setAttributes({ [key]: value }, { skipWatcherUpdates: false });
```
artf: `setAttributes` overwrites existing attributes, shouldn't we use `addAttributes`?

packages/core/test/specs/data_sources/model/conditional_variables/__snapshots__/ConditionalTraits.ts.snap
```diff
@@ -14,17 +14,30 @@ exports[`TraitConditionalVariable should store traits with conditional values co
...
+                    "changeProp": false,
+                    "default": "",
+                    "id": "data-variable-id",
+                    "label": "",
+                    "name": "dynamicTrait",
+                    "options": [],
+                    "placeholder": "",
+                    "step": 1,
+                    "type": "text",
+                    "unit": "",
```
artf: Seems like we're now storing the trait definition here, I think that might be a problem.

packages/core/src/dom_components/model/ComponentDynamicValueListener.ts
```diff
@@ -0,0 +1,123 @@
...
+  private attributeWatcher: DynamicValueWatcher;
+  private traitsWatcher: DynamicValueWatcher;
+
+  constructor(
+    private component: Component,
+    em: EditorModel,
+  ) {
+    this.propertyWatcher = new DynamicValueWatcher(this.createPropertyUpdater(), em);
+    this.attributeWatcher = new DynamicValueWatcher(this.createAttributeUpdater(), em);
+    this.traitsWatcher = new DynamicValueWatcher(this.createTraitUpdater(), em);
```
artf: You know, I have some doubts regarding how we deal with traits here and inside Component.
I'd expect traits to read all the values directly from props/attributes based on the trait definition.

packages/core/src/dom_components/model/Component.ts
```diff
@@ -1554,11 +1570,14 @@ export default class Component extends StyleableModel<ComponentProperties> {
    * @private
    */
   toJSON(opts: ObjectAny = {}): ComponentDefinition {
-    const obj = Model.prototype.toJSON.call(this, opts);
-    obj.attributes = this.getAttributes();
+    let obj = Model.prototype.toJSON.call(this, opts);
+    obj = { ...obj, ...this.componentDVListener.getDynamicPropsDefs() };
+    obj.attributes = this.componentDVListener.getAttributesDefsOrValues(this.getAttributes({ noClass: true }));
+    obj[dynamicAttrKey] = this.serializeDynamicTraits();
```
artf: not sure if we need to store any kind of trait information here 🤔

packages/core/test/specs/data_sources/dynamic_values/attributes.ts
```diff
@@ -0,0 +1,247 @@
+import Editor from '../../../../src/editor/model/Editor';
+import DataSourceManager from '../../../../src/data_sources';
+import ComponentWrapper from '../../../../src/dom_components/model/ComponentWrapper';
+import { DataVariableType } from '../../../../src/data_sources/model/DataVariable';
+import { setupTestEditor } from '../../../common';
+import { Component } from '../../../../src';
+
+const staticAttributeValue = 'some tiltle';
+describe('TraitDataVariable', () => {
```
artf: let's update the name of the test

packages/core/src/dom_components/model/ComponentDynamicValueListener.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  private createPropertyUpdater() {
+    return (key: string, value: any) => {
+      this.component.set(key, value, { skipWatcherUpdates: false });
```
artf: I think when we update the component from data sources changes, we don't want this change to be counted as a dirty change and skip it from the undoManager (`avoidStore` option should already handle both cases).

Probably DataSource itself could define if it needs to trigger the dirty change and if the change is undoable.
For example, a data source with external records (eg. REST API) could retrigger the fetch of all records on a specific interval. The editor would update components with new data but I'd not expect changes on the project data itself.
In case of static data source records (records stored in the project JSON), their update will trigger the dirty change and could be defined as an undoable action.

packages/core/src/dom_components/model/Component.ts
```diff
@@ -289,7 +295,7 @@ export default class Component extends StyleableModel<ComponentProperties> {
     this.opt = opt;
     this.em = em!;
     this.config = opt.config || {};
-    this.set('attributes', {
+    this.setAttributes({
       ...(result(this, 'defaults').attributes || {}),
       ...(this.get('attributes') || {}),
     });
```
mohamedsalem401, line 301: This's needed to convert dynamic definitions to static values

artf (approved):
Awesome 💜

PR #6359, opened 2024-12-23
Add collection components

mohamedsalem401:


packages/core/src/dom_components/model/Component.ts
```diff
@@ -292,12 +298,15 @@ export default class Component extends StyleableModel<ComponentProperties> {
     }
 
     opt.em = em;
-    this.opt = opt;
+    this.opt = {
+      ...opt,
+      collectionsStateMap: props[keyCollectionsStateMap],
+      isCollectionItem: !!props['isCollectionItem'],
```
artf: any reason to store these in `opt`? I'd expect to get them from the model if I need them

packages/core/src/dom_components/model/Component.ts
```diff
@@ -1591,6 +1605,17 @@ export default class Component extends StyleableModel<ComponentProperties> {
     delete obj.open; // used in Layers
     delete obj._undoexc;
     delete obj.delegate;
+    if (this.get('isCollectionItem')) {
```
artf: keyIsCollectionItem?

packages/core/src/dom_components/model/Component.ts
```diff
@@ -1591,6 +1605,17 @@ export default class Component extends StyleableModel<ComponentProperties> {
...
+      delete obj[keySymbol];
+      delete obj[keySymbolOvrd];
+      delete obj[keySymbols];
+      delete obj[keyCollectionsStateMap];
+      delete obj['isCollectionItem'];
+      delete obj.attributes.id;
+      obj['components'] = this.components()
+        .toArray()
+        .map((cmp) => cmp.toJSON());
```
artf: Any reason for this? I'd expect the parent call `Model.prototype.toJSON.call(this, opts)` to do already the same

packages/core/src/dom_components/model/Component.ts
```diff
@@ -1657,9 +1682,7 @@ export default class Component extends StyleableModel<ComponentProperties> {
    * @return {this}
    */
   setId(id: string, opts?: SetOptions & { idUpdate?: boolean }) {
-    const attrs = { ...this.get('attributes') };
-    attrs.id = id;
-    this.set('attributes', attrs, opts);
+    this.addAttributes({ id });
```
artf: don't forget `opts` here

packages/core/src/dom_components/model/ComponentDynamicValueWatcher.ts
```diff
@@ -1,46 +1,88 @@
...
+    return evaluatedProps;
   }
 
   removeAttributes(attributes: string[]) {
     this.attributeWatcher.removeListeners(attributes);
+    this.updateSymbolOverride();
+  }
+
+  private updateSymbolOverride() {
+    if (!this.component || !this.component.get('isCollectionItem')) return;
```
artf: keyIsCollectionItem

packages/core/src/utils/sorter/CanvasComponentNode.ts
```diff
@@ -6,6 +6,17 @@ export default class CanvasComponentNode extends BaseComponentNode {
...
+   * Check if a source node can be moved to a specified index within this component.
+   * @param {BaseComponentNode} source - The source node to move.
+   * @param {number} index - The display index to move the source to.
+   * @returns {boolean} - True if the move is allowed, false otherwise.
+   */
+  canMove(source: BaseComponentNode, index: number): boolean {
+    console.log('🚀 ~ CanvasComponentNode ~ canMove ~ this.model:', this.model);
+    console.log('🚀 ~ CanvasComponentNode ~ canMove ~ source.model:', source.model);
+    return this.model.em.Components.canMove(this.model, source.model, this.getRealIndex(index)).result;
+  }
```
artf: do we need this or you're using it just for debugging?

packages/core/src/data_sources/model/collection_component/types.ts
artf, line 1: * update keys to camelCase
* Let's use interfaces instead of types when possible

packages/core/src/data_sources/model/collection_component/types.ts
```diff
@@ -0,0 +1,47 @@
...
+export type CollectionComponentDefinition = {
+  [keyCollectionDefinition]: CollectionDefinition;
+} & ComponentDefinition;
+
+export type CollectionDefinition = {
+  type: typeof CollectionComponentType;
+  collection_name?: string;
+  config: CollectionConfig;
+  block: ComponentDefinition;
+};
```
artf: I'm not sure about this interface. 
* Let's avoid naming confusion (eg. backbone collections). We have `DataVariable*`, `DataCondition*`, so let's call it `DataCollection*`
* Why do we need `collection_name`? I would expect DataCollection to ask mainly `dataSourceId`, if there is a name, that should get the one directly from the connected dataSource.
* `block` seems too much related to GrapesJS Blocks. I'd call it `componentDef`
* Do you think we need `config`? Would it be easier to have these configs directly in the definition itself?!

packages/core/test/specs/data_sources/model/collection_component/CollectionComponent.ts
```diff
@@ -0,0 +1,863 @@
...
+                    variable_type: CollectionStateVariableType.current_item,
+                    path: 'user',
+                  },
+                },
+              ],
+              content: {
+                type: CollectionVariableType,
+                variable_type: CollectionStateVariableType.current_item,
+                path: 'user',
+              },
```
artf: Let's avoid using `componentDef.content`, it's a bit of edge–case thing (custom RTE). 
The goal of `content` is to get/render the HTML as it is, without parsing it to components. So it doesn't even make sense to have a component with both `componentDef.content` and `componentDef.components`.

packages/core/test/specs/data_sources/model/collection_component/CollectionComponent.ts
```diff
@@ -0,0 +1,863 @@
...
+            },
+            config: {
+              dataSource: {
+                type: DataVariableType,
+                path: 'my_data_source_id',
+              },
+            },
+          },
+        })[0];
```
artf: I'm trying some examples locally and I don't see it working as expected.

This seems to work
```js
{
  type: 'collection-component',
  style: { display: 'flex', gap: '10px', 'flex-direction': 'column' },
  collectionDefinition: {
    block: {
      style: { display: 'flex', gap: '10px' },
      content: {
        type: 'parent-collection-variable',
        variable_type: 'current_item',
        path: 'user',
      },
    },
    config: {
      dataSource: {
        type: 'data-variable',
        path: 'my_data_source_id',
      },
    },
  },
}
```
This doesn't (probably a more common use case)
```js
{
  type: 'collection-component',
  style: { display: 'flex', gap: '10px', 'flex-direction': 'column' },
  collectionDefinition: {
    block: {
      style: { display: 'flex', gap: '10px' },
      components: [
        {
          tagName: 'span',
          components: {
            type: 'parent-collection-variable',
            variable_type: 'current_item',
            path: 'user',
          }
        },
        {
          tagName: 'custom-el',
          components: {
            type: 'parent-collection-variable',
            variable_type: 'current_item',
            path: 'age',
          }
        },
      ],
    },
    config: {
      dataSource: {
        type: 'data-variable',
        path: 'my_data_source_id',
      },
    },
  },
}
```

Also, do we really need the new `CollectionVariableType`? Looks like then we'll also be forced to create a new type for conditions.
As you're already adding some kind of differentiator (`variableType`), can't we simply reuse `DataVariableType`?

packages/core/src/utils/sorter/CanvasComponentNode.ts
mohamedsalem401: Forgot to delete it while testing drag and drop behavior for collection items 😅

artf (approved):

Issue #1234, opened 2018-06-25
Template Problems with GrapesJS


ivanezeigbo:
1. Some templates just don't load well. Very simple templates do load properly, but there seems to be a problem with templates with more functions and css. For example, loading a template like [this](http://webbuild.jenezys.com/templates/free_templates/aerial/index.html) would generate some problems. First, the animation fade in of the letters would not occur. Once inside the #gjs div tags, the template gets stuck in the initial loading and does not completely load even with allowScripts. And it has only two css files linked and a simple js code. If the texts and icons happen to load by chance at some point, hovering over them makes them disappear again because there is a hover function attached to them. You may also check out [this other template](http://webbuild.jenezys.com/templates/free_templates/Kreo10/index.html), which is more complicated. The editor just gets stuck at the loader and would not be able to show the rest of the content. So it becomes impossible to edit such templates. 

2. Also, is it possible to have an "unlink" tooltip or something that allows us to unlink images that are linked? One cannot simply carry the image out of the link block like you can do with the editor's link block. So perhaps, this would allow one to delete the <a> tags surrounding the image given that the editor may not recognize this as a link block.

ivanezeigbo:
@artf I don't understand. You tagged this as invalid.

artf:
Follow the issue template please and always write ONE problem/question/bug per issue

PR #6367, opened 2025-01-10
Modify Chinese translation 

Renzo-Lee:
New translation #1234

### DIFF
--- a/packages/core/src/i18n/locale/zh.js
+++ b/packages/core/src/i18n/locale/zh.js
@@ -1,86 +1,86 @@
 const traitInputAttr = { placeholder: '例子. 输入文字' };
 
 export default {
-  add: "添加",
-  delete: "删除",
-  duplicate: "创建副本",
-  rename: "重命名",
-  remove: "移除",
-  clear: "清空",
-  select: "选择",
-  selectList: "从列表中选择",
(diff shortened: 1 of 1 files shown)


artf:
Hi @Renzo-Lee, it seems like there are a lot of properties not part of the core labels, from where did you get them?

Renzo-Lee:
> Hi @Renzo-Lee, it seems like there are a lot of properties not part of the core labels, from where did you get them?

I discovered the complete en language data while using StudioEditor

微信图片_20250115182945
微信图片_20250115182950

artf:
oh yeah, but that's not part of the core 🥲
Maybe one day we'll expose the Studio locales files but right now it's wrong to push them here

Renzo-Lee:
> oh yeah, but that's not part of the core 🥲 Maybe one day we'll expose the Studio locales files but right now it's wrong to push them here

Ok, the en file in [packages/core/src/i18n/locality] is not comprehensive enough. I would like to translate the language into Chinese. Is there any other place to refer to?

artf:
@Renzo-Lee we don't yet have Studio-related local files exposed but we'll do it one day. For now, you can keep those files translated in your own repo but we can't merge them here as those keys are not related to the core.

PR #6382, opened 2025-01-22
Fix resizer on absolute drag mode when the parent is positioned ( position: relative | absolute | .... )

### LINKED ISSUE #6373: BUG: Resizing of Child with absolute drag enabled positions element wrongly when in Parent with relative position
### GrapesJS version

Version 131.0.6778.265

### Reproducible demo link

https://codepen.io/chipikaf/pen/KwPexjL

### Describe the bug

**How to reproduce the bug?**
1. Drop the button component in the Middle section
2. Move the button anywhere in the section
3. Resize element

For the element to be resized in its original position

The element is moved below to an unexpected position

### Code of Conduct

mohamedsalem401:
For https://github.com/GrapesJS/grapesjs/issues/6373


packages/core/src/utils/Resizer.ts
```diff
@@ -750,4 +756,20 @@ export default class Resizer {
...
+    while (currentElement) {
+      const position = window.getComputedStyle(currentElement).position;
+      if (position === 'relative' || position === 'absolute' || position === 'fixed' || position === 'sticky') {
+        return true;
+      }
+      currentElement = currentElement.parentElement;
+    }
+
+    return false;
+  }
```
artf: this looks expensive, can't we simply rely on [`offsetParent`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent)?

artf (approved):

PR #6400, opened 2025-02-07
fix: typos in documentation files

maximevtush:


artf (commented):
thank you @maximevtush one note though, `docs/api/*` files are auto generated so you would need to update their source files

artf (approved):

PR #6426, opened 2025-02-27
Remove deprecated @import commands from SASS files

sirbeagle:
This PR is to address the deprecation of the `@import` in the SASS processor we are using - details are [here](https://sass-lang.com/documentation/at-rules/import/).

The idea here is to remove the use of `@import` statements in sass files and remove nested calls that assume a parent style is available and instead reference shared parent items with `@use`.

I used the [sass-migrator](https://sass-lang.com/documentation/cli/migrator/) tool as a starting point and then did a bunch of refactoring of how the files are being called. I did break some items into new files (`gjs_main_mixins.scss`, `gjs_category_general.scss`, etc) so that other `_gjs` files could `@use` them to do things like extending a style.

While I did move stuff around a bit, I did my best not to touch the content or order of styles unless if was absolutely necessary. Obviously these changes touch a core component of the library so we'll need to test the heck out of it if we decide to proceed.


sirbeagle:
I have built the compiled CSS files both with the existing `dev` branch and with this PR and compared them using [cssdiff](https://inlife.github.io/cssdiff/#). The resulting files are about identical, the difference being that sometimes when defining multiple rules (separated by comma) the order may be different even though the rule is the same:

image

As far as I can tell this doesn't affect the styles at all, and it only is happening when multiple selectors are defined like above. I'm still checking everything over but I think this might be ready to go shortly.

sirbeagle:
I have gotten the output `grapes.min.css` to produce the same content as the `dev` branch with the exception of the order changes of multiple selector definitions mentioned in the comment above (the red / green image). I think this might be ready to go.

artf (approved):
Checked locally and looks all good, awesome work 👏

PR #6451, opened 2025-03-17
Dynamic values improvements

mohamedsalem401:
- Refactored code and fixed bugs.
- Added flexible conditional component display, allowing all components to render dynamically based on conditions.


packages/core/src/data_sources/model/ComponentDataVariable.ts
```diff
@@ -17,10 +17,21 @@ export default class ComponentDataVariable extends Component {
     };
   }
 
-  constructor(props: DataVariableProps, opt: ComponentOptions) {
-    super(props, opt);
-    const { type, path, defaultValue } = props;
+  constructor({ type, path, defaultValue }: DataVariableProps, opt: ComponentOptions) {
+    super({ type, path, defaultValue }, opt);
```
artf: This is a Component, as a component I should be able to pass `tagName`, `attributes` and all other model specific properties.

I'd suggest a new property for components with data resolvers.
```js
get defaults() {
    return {
      // @ts-ignore
      ...super.defaults,
      droppable: false,
      type: DataVariableType,
      dataResolver: {
        path: '',
        defaultValue: '',
      }
    };
  }
```

packages/core/src/data_sources/model/ComponentDataVariable.ts
```diff
@@ -17,10 +17,21 @@ export default class ComponentDataVariable extends Component {
...
+    this.setupChangeListeners();
+  }
+
+  private setupChangeListeners() {
+    const handleChange = (property: keyof DataVariableProps) => (component: Component, newValue: string) => {
+      this.dataResolver.set(property, newValue);
+    };
+
+    this.listenTo(this, 'change:path', handleChange('path'));
+    this.listenTo(this, 'change:defaultValue', handleChange('defaultValue'));
```
artf: For model listeners `this.on('...'` should be fine.

Maybe this could also be managed with one listener?! Then we could do this based on the suggested prop above:
```js
this.dataResolver.set(this.get('dataResolver'));
```

packages/core/src/data_sources/model/conditional_variables/ComponentDataCondition.ts
```diff
@@ -1,61 +1,210 @@
...
+      'syncChildState',
+      'initIfTrueComponents',
+      'initIfFalseComponents',
+      'initOutputComponents',
+      'initComponentsByType',
     );
-    this.dataResolver = dataConditionInstance;
-    this.dataResolver.onValueChange = this.handleConditionChange.bind(this);
+
+    this.dataResolver = new DataCondition(updatedProps, { em: opt.em });
```
artf: `new DataCondition` should only receive related props to it (no component-related props, eg. `tagName`)

packages/core/src/data_sources/model/conditional_variables/ComponentDataCondition.ts
```diff
@@ -1,61 +1,210 @@
...
+
+    this.initIfTrueComponents();
+    this.initIfFalseComponents();
+    this.initOutputComponents();
+    this.listenTo(this.components(), 'add', this.syncChildState);
+    this.dataResolver.onValueChange = this.initOutputComponents;
```
artf: maybe let's make this an option in DataCondition?

packages/core/src/data_sources/model/conditional_variables/ComponentDataCondition.ts
```diff
@@ -1,61 +1,210 @@
...
+      ifFalse: ifFalse,
+    } as any;
+    super(updatedProps, opt);
+    bindAll(
+      this,
+      'syncChildState',
+      'initIfTrueComponents',
+      'initIfFalseComponents',
+      'initOutputComponents',
+      'initComponentsByType',
```
artf: are you sure you need all of these?
eg. I don't see why you would do it for something like `initIfTrueComponents`

packages/core/src/data_sources/model/conditional_variables/DataCondition.ts
```diff
@@ -39,82 +40,117 @@ export class DataCondition extends Model<DataConditionPropsDefined> {
...
   }
 
-  setIfTrue(newIfTrue: any) {
-    this.set('ifTrue', newIfTrue);
+  private ensureConditionEvaluatorInstance() {
+    const condition = this.get('condition') ?? {};
+    if (condition instanceof DataConditionEvaluator) return condition;
+
+    const instance = new DataConditionEvaluator({ condition }, { em: this.em });
+    this.set('condition', instance, { silent: true });
```
artf: would it make sense to store it at the model instance level?
```js
this._conditionEvaluator = instance
```

packages/core/src/data_sources/model/ComponentDataVariable.ts
```diff
@@ -40,11 +51,11 @@ export default class ComponentDataVariable extends Component {
   }
 
   setPath(newPath: string) {
-    this.dataResolver.set('path', newPath);
+    this.set('path', newPath);
   }
 
   setDefaultValue(newValue: string) {
-    this.dataResolver.set('defaultValue', newValue);
+    this.set('defaultValue', newValue);
```
artf: let's write directly to `this.dataResolver`

packages/core/src/data_sources/model/conditional_variables/ComponentDataCondition.ts
artf, line 1: I think we have to rethink about the logic here.

In order to help yourself with the proper architecture, try to always think about, how would you "import" something like this from an HTML code?

IMHO this might be the right expectation:

```html
<mytag data-gjs-type="data-condition" data-gjs-dataResolver='{ "condition": {...} }' >
  <div data-gjs-type="data-condition-true">True <b>component</b></div>
  <div data-gjs-type="data-condition-false">
   <div data-gjs-type="data-condition" data-gjs-dataResolver='{ "condition": {...} }' >
     <div data-gjs-type="data-condition-true">Nested true component</div>
     <div data-gjs-type="data-condition-false">Nested false</div>
   </div>
  </div>
</mytag>
```
On the Component Model:
```js
{
  type: 'data-condition',
  tagName: 'mytag',
  dataResolver: { condition: {...} },
  components: [
   {   type: 'data-condition-true', components: [....] },
   {   type: 'data-condition-false', components: [....] },
  ]
}
```

On the Component View, by default, you switch what component model to render based on the condition listener (to subscribe on the View level).

It's ok to have `ifTrue`/`ifFalse` in the model defintion but those should only be used for component initialization. If you have already `data-condition-true`/`data-condition-false` inside the component/componentDef, we should keep/use those.

Another important thing to keep in mind, I'd not expect model updates when the condition changes (eg. when you change the model, that updates the state of the document, it triggers storage/undoManager, etc.)

packages/core/src/data_sources/model/conditional_variables/ComponentDataCondition.ts
```diff
@@ -1,61 +1,130 @@
...
+    const components = this.components();
+    const existingComponent = components.at(index);
+
+    if (existingComponent) {
+      this.components().remove(existingComponent);
+      const newComponents = this.append(newContent, { at: index })[0];
+      this.trigger(DataConditionOutputChangedEvent);
+      return newComponents;
+    }
+  }
```
artf: Not sure about this. I think we should not allow removing the already defined true/false.
I think you just need to update those components.
```js
existingComponent?.components(newContent);
```

packages/core/src/data_sources/model/conditional_variables/ComponentDataCondition.ts
```diff
@@ -1,61 +1,130 @@
...
+      this.dataResolver.set(this.get('dataResolver'));
+    });
+  }
+
+  toJSON(opts?: ObjectAny): ComponentProperties {
+    const json = super.toJSON(opts);
+    const dataResolver = this.dataResolver.toJSON();
+    delete dataResolver.type;
+    delete dataResolver.ifTrue;
+    delete dataResolver.ifFalse;
```
artf, line 112: So, we don't really use these `ifTrue`/`ifFalse` anymore, correct?

Is it just because of how conceptually DataCondition works?!

packages/core/src/data_sources/view/ComponentDataConditionView.ts
```diff
@@ -1,4 +1,62 @@
...
+    }
+  }
+
+  postRender() {
+    this.renderDataResolver();
+    super.postRender();
+  }
+
+  remove() {
+    this.stopListening(this.model.components(), 'reset', this.postRender.bind(this));
```
artf: be careful, bind() returns a new function, so the `stopListening` has no effect in this case

packages/core/src/data_sources/view/ComponentDataConditionView.ts
```diff
@@ -1,4 +1,62 @@
...
+    return this.model.getOutputContent();
+  }
+
+  getIfTrueContent() {
+    return this.model.getIfTrueContent();
+  }
+
+  getIfFalseContent() {
+    return this.model.getIfFalseContent();
+  }
```
artf: I'd say these methods are not necessary

artf (commented):
🚀

artf (approved):
Awesome 💜

Issue #2721, opened 2020-04-17
[Bug] Can not import any locale file from grapesjs/locale/


alex1989:
Hi I run into an error when I'm trying to import any locale file from 'grapejs/locale' by using es6 import

I followed by the documentation:

https://grapesjs.com/docs/modules/I18n.html#configuration

but I always get this:
![image](https://user-images.githubusercontent.com/2537532/79566144-a0fc2780-80e4-11ea-9915-44f2c5a9925f.png)

I can import it only in next way
![image](https://user-images.githubusercontent.com/2537532/79565085-83c65980-80e2-11ea-9bbe-32f0dc36a4cd.png)

Could you explain why I can't import it directly from `grapesjs/locale/`.

grapejs was installed by npm install

grapejs is latest version (0.16.3)

artf:
Should be fixed in https://github.com/artf/grapesjs/releases/tag/v0.16.12

artf:
@willGabrielPereira can you try with https://github.com/GrapesJS/grapesjs/releases/tag/v0.20.4

teknosains:
> @willGabrielPereira can you try with https://github.com/GrapesJS/grapesjs/releases/tag/v0.20.4

im using **[v0.22.3](https://github.com/GrapesJS/grapesjs/releases/tag/v0.22.3)**, still doesnt work

teknosains:
anyone have a work around?

padcom:
I'm using 0.22.6 and the problem is still not fixed.

I am using Vite and bundling all together, including locales does work. However, if you'd like to import `grapesjs/locale/pt` using 

```html
  <script type="importmap">
    {
      "imports": {
        "grapesjs": "https://unpkg.com/grapesjs/dist/grapes.mjs",
        "grapesjs/locale/en": "https://unpkg.com/grapesjs/locale/en.js",
        "grapesjs/locale/pt": "https://unpkg.com/grapesjs/locale/pt.js",
      }
    }
  </script>
```

it won't work:

```
Uncaught SyntaxError: The requested module 'grapesjs/locale/pt' does not provide an export named 'default'
```

I believe the problem occurs because the locales are not in ECMAScript module format but are geared towards bundlers:

```javascript
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports.default = void 0;
var traitInputAttr = {
  placeholder: 'Wpisz tutaj'
};
var _default = exports.default = {
  assetManager: {
[...]

PR #6459, opened 2025-03-21
Add ECMAScript locale modules

padcom:
Implements a possible workaround for #2721 

What's happening during the build is that first the source locale files are copied over from `packages/core/src/locale` to `packages/core/locale` and then pushed through webpack for transpilation and the original files are overridden. 

For use cases where an ECMAScript module is required (e.g. when importing files using the `<script type="importmap">`) it is imperative that the files are not transpiled. For that reason, before the transpilation, the original content is copied over to `<filename>.mjs` thus making a copy of the file in ECMAScript module format, which those files are originally in.

### DIFF
--- a/packages/cli/src/build.ts
+++ b/packages/cli/src/build.ts
@@ -52,9 +52,14 @@ export const buildLocale = async (opts: BuildOptions = {}) => {
   const babelOpts = { ...babelConfig(buildWebpackArgs(opts) as any) };
   fs.readdirSync(localDst).forEach((file) => {
     const filePath = `${localDst}/${file}`;
+    const esModuleFileName = filePath.replace(/\.[^.]+$/, '.mjs');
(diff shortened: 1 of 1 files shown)


padcom:
With this change it is now possible to do the following:

```html
  <script type="importmap">
    {
      "imports": {
        "grapesjs": "https://unpkg.com/grapesjs/dist/grapes.mjs",
        "grapesjs/locale/en": "https://unpkg.com/grapesjs/locale/en.mjs",
        "grapesjs/locale/pt": "https://unpkg.com/grapesjs/locale/pt.mjs",
      }
    }
  </script>
```

Make sure you also load the styles correctly:

```html
  <link rel="stylesheet" href="https://unpkg.com/grapesjs/dist/css/grapes.min.css">
```

artf (approved):
That's awesome 💜

PR #6471, opened 2025-04-01
Adjust ComponentDrag types

itscarlosrufo:


packages/core/src/commands/index.ts
```diff
@@ -450,4 +451,9 @@ export default class CommandsModule extends Module<CommandsConfig & { pStylePref
     this.commands = {};
     this.active = {};
   }
+
+  // Dummy property to ensure ComponentDragEventProps type is included in the build
+  dummyComponentDragEventProps() {
+    return undefined as ComponentDragEventProps | undefined;
+  }
```
itscarlosrufo: @artf this is the 'cleanest' way I found to get `ComponentDragEventProps` into the build types. Only importing/exporting doesn't work due the tree-shaking, but wondering if you've got another idea to get that type in the build without modifying the logic

packages/core/src/commands/view/ComponentDrag.ts
```diff
@@ -127,7 +125,7 @@ export default {
   },
 
   getGuidesStatic() {
-    let result: Guide[] = [];
+    let result: ComponentGuide[] = [];
```
itscarlosrufo: I had to change the name because when generating the build, I found out there's already a Guide type, now it's unique

packages/core/src/commands/index.ts
itscarlosrufo: Fixed on 84ef292, exporting the type in `src/index` did the trick

packages/core/src/commands/view/ComponentDrag.ts
artf: Yeah, correct, all TS types must be unique as everything is exported in a single file.

I'm pushing a rename to `ComponentDragGuide` as it gives a bit more context.

artf (approved):
🚀

PR #6482, opened 2025-04-08
Dynamic values improvements

mohamedsalem401:
- Remove DataCollectionVariable ( merge it with DataVariable )
- Update API for ComponentDataCollection
- Update ComponentDataVariable
- Add some method to DataResolvers


packages/core/src/data_sources/model/data_collection/ComponentDataCollection.ts
```diff
@@ -187,57 +206,75 @@ export default class ComponentDataCollection extends Component {
...
     const json = super.toJSON.call(this, opts) as ComponentDataCollectionProps;
-    json[keyCollectionDefinition].componentDef = this.getFirstChildJSON();
-    delete json.components;
     delete json.droppable;
-    return json;
+    delete json[keySymbol];
+    delete json.attributes?.id;
+
+    const components = new Components(this.components().at(0), this.opt) as any;
+    return { ...json, components };
```
artf: any reason to recreate `Components`?

packages/core/src/data_sources/model/data_collection/ComponentDataCollection.ts
```diff
@@ -20,10 +25,25 @@ import {
...
+      // @ts-ignore
+      ...super.defaults,
+      droppable: false,
+      type: DataCollectionType,
+      components: [
+        {
+          type: DataCollectionItemType,
+        },
+      ],
+    };
```
artf, line 44: don't we have `dataResolver` configuration here as the variable and condition?

eg. we should be able to import collections config from HTML

packages/core/src/data_sources/model/data_collection/ComponentDataCollection.ts
```diff
@@ -155,25 +166,33 @@ export default class ComponentDataCollection extends Component {
     this.components().reset(this.getCollectionItems(), updateFromWatcher as any);
   }
 
-  getCollectionItems() {
-    const { componentDef, collectionConfig } = this.collectionDef;
-    const result = validateCollectionConfig(collectionConfig, componentDef, this.em);
+  private getCollectionItems() {
+    const firstChild = this.ensureFirstChild();
+    // TODO: Move to component view
+    firstChild.addStyle({ display: 'none' });
```
artf: can we move it now? Looks a bit wrong here

packages/core/src/data_sources/model/data_collection/ComponentDataCollection.ts
```diff
@@ -187,57 +206,75 @@ export default class ComponentDataCollection extends Component {
...
-          {
-            ...componentDef,
-            draggable: false,
-            removable: false,
-          },
-          this.opt,
-        );
+        setCollectionStateMapAndPropagate(firstChild, collectionsStateMap, collectionId);
+        // TODO: Move to component view
+        firstChild.addStyle({ display: '' });
```
artf: same as above

packages/core/src/data_sources/model/data_collection/ComponentDataCollection.ts
```diff
@@ -187,57 +206,75 @@ export default class ComponentDataCollection extends Component {
...
+
+        continue;
       }
 
-      const instance = symbolMain!.clone({ symbol: true });
-      !isFirstItem && instance.set('locked', true);
-      setCollectionStateMapAndPropagate(collectionsStateMap, collectionId)(instance);
+      const instance = firstChild!.clone({ symbol: true });
+      instance.set('locked', true);
```
artf: should we use avoidStore/partial here?

packages/core/src/data_sources/model/data_collection/ComponentDataCollection.ts
```diff
@@ -95,25 +114,26 @@ export default class ComponentDataCollection extends Component {
...
   setDataSource(dataSource: DataCollectionDataSource) {
     this.set(keyCollectionDefinition, {
       ...this.collectionDef,
-      collectionConfig: { ...this.collectionConfig, dataSource },
+      dataSource,
     });
   }
 
+  setCollectionItemComponents(content: ComponentAddType) {
+    this.components().at(0).components(content);
```
artf: let's create a getter for `this.components().at(0)`?

packages/core/test/specs/data_sources/model/data_collection/ComponentDataCollection.getters-setters.ts
```diff
@@ -40,37 +39,40 @@ describe('Collection component getters and setters', () => {
...
-          collectionConfig: {
-            collectionId: 'my_collection',
-            startIndex: 1,
-            endIndex: 2,
-            dataSource: {
-              type: DataVariableType,
-              path: 'my_data_source_id',
-            },
+        },
+        collectionDef: {
```
artf: I guess `collectionDef` should be `dataResolver` to be consistent with variable/condition

artf (commented):
left a few comments but overall looks amazing

packages/core/src/data_sources/model/data_collection/ComponentDataCollection.ts
mohamedsalem401: We do, so I'll update this method, but it's called `componentDef` here, so do you think we should update it to `dataResolver` instead?

```ts
get defaults(): ComponentDefinitionDefined {
    return {
      // @ts-ignore
      ...super.defaults,
      droppable: false,
      type: DataCollectionType,
      components: [
        {
          type: DataCollectionItemType,
        },
      ],
      [keyCollectionDefinition]: {
        collectionId: '',
        startIndex: 0,
        endIndex: Number.MAX_VALUE,
        dataSource: {
          type: DataVariableType,
          path: '',
        },
      },
    };
  }
```

packages/core/src/data_sources/model/data_collection/ComponentDataCollection.ts
mohamedsalem401: Because we want only to have the first component exported 🥲. This will only be here until we move depending on `Components` to display the items in the `View`. let me know wdt

packages/core/src/data_sources/model/data_collection/ComponentDataCollection.ts
artf: Yeah it will be more consistent with other data-related components

packages/core/src/data_sources/model/data_collection/ComponentDataCollection.ts
artf: right, but you could do this: `return { ...json, components: [this.components().at(0)] };`
I guess it will be the same

PR #6483, opened 2025-04-09
Absolute mode enhancements

itscarlosrufo:


packages/core/src/dom_components/model/Component.ts
```diff
@@ -1146,7 +1146,21 @@ export default class Component extends StyleableModel<ComponentProperties> {
...
+              const clientX = rect ? rect.left + rect.width / 2 : 0;
+              const clientY = rect ? rect.top + rect.height / 2 : 0;
+
+              // tld-move event needs to be called with a target and a drag event
+              ed.runCommand('tlb-move', {
+                target,
+                event: new DragEvent('dragstart', { clientX, clientY }),
+              });
+            });
+          },
```
itscarlosrufo: Multiple selection doesn't work with this code, but individual selection it's working. We can merge it like this, and improve in a follow-up PR (or I can take a look tomorrow)

artf (approved):

PR #6501, opened 2025-04-24
Add an option to convert attributes with hyphens to camel case 

mohamedsalem401:
```ts

  /**
   * Configures whether `data-gjs-*` attributes should be automatically converted from hyphenated to camelCase.
   *
   * When `true`:
   * - Hyphenated `data-gjs-*` attributes (e.g., `data-gjs-my-component`) are transformed into camelCase (`data-gjs-myComponent`).
   * - If `defaults` contains the camelCase version and not the original attribute, the camelCase will be used; otherwise, the original name is kept.
   *
   * @default false
   */
  convertDataGjsAttributesHyphens?: boolean;
```


artf (approved):
🚀

PR #6529, opened 2025-05-27
Datasources improvements

mohamedsalem401:
- Fix bug with dynamic components having `dataResolver` with `type: 'data-variable'`
- Allow collections to loop over objects
- Add option to paste datasource values as plainText


artf (approved):
👍

PR #6531, opened 2025-06-04
Update component styles

mohamedsalem401:
- Refactor how component styles work ( make them like properties and attributes )


artf (approved):
💜

PR #6553, opened 2025-06-27
fix drag and drop

mohamedsalem401:
This PR fixes an issue where components dragged from an undroppable region into a droppable one were not being recognized. The fix implements a check that detects the transition based on the last mouse event, allowing the sorter to correctly handle these movements.

### DIFF
--- a/packages/core/src/utils/sorter/DropLocationDeterminer.ts
+++ b/packages/core/src/utils/sorter/DropLocationDeterminer.ts
@@ -117,6 +117,7 @@ export class DropLocationDeterminer<T, NodeType extends SortableTreeNode<T>> ext
     this.adjustForScroll();
     const { targetNode: lastTargetNode } = this.lastMoveData;
     this.eventHandlers.onMouseMove?.(mouseEvent);
+    this.cacheContainerPosition();
     const { mouseXRelative: mouseX, mouseYRelative: mouseY } = this.getMousePositionRelativeToContainer(
       mouseEvent.clientX,
       mouseEvent.clientY,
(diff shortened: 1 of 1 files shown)


packages/core/src/utils/sorter/DropLocationDeterminer.ts
```diff
@@ -279,17 +280,6 @@ export class DropLocationDeterminer<T, NodeType extends SortableTreeNode<T>> ext
...
-   * Checks if the target node has changed and returns the last one if they are identical.
-   *
-   * @param targetNode - The newly calculated target node.
-   * @returns The new or reused target node.
-   */
-  private getOrReuseTargetNode(targetNode?: NodeType): NodeType | undefined {
-    const lastTargetNode = this.lastMoveData.targetNode;
-    return targetNode?.equals(lastTargetNode) ? lastTargetNode : targetNode;
-  }
-
```
mohamedsalem401, line 292: This has nothing to do with the PR, just cleanup

packages/core/src/utils/sorter/DropLocationDeterminer.ts
```diff
@@ -382,16 +372,22 @@ export class DropLocationDeterminer<T, NodeType extends SortableTreeNode<T>> ext
...
     const sameHoveredNode = targetNode.equals(lastHoveredNode);
     targetNode.nodeDimensions = sameHoveredNode ? lastHoveredNode!.nodeDimensions! : this.getDim(targetNode.element!);
     const hoverIndex = this.getIndexInParent(targetNode, targetNode.nodeDimensions!, mouseX, mouseY);
     const sameHoveredIndex = hoverIndex === lastHoveredIndex;
-    const sameHoverPosition = sameHoveredNode && sameHoveredIndex;
+    const isWithinDropArea = targetNode.isWithinDropBounds(mouseX, mouseY);
+    const sameHoverPosition =
+      sameHoveredNode &&
+      sameHoveredIndex &&
+      isWithinDropArea === targetNode.isWithinDropBounds(lastMouseEvent?.clientX ?? 0, lastMouseEvent?.clientY ?? 0);
```
mohamedsalem401, line 386: Here, if last movement was over an undraggable area but now it's on draggable area, don't use the cached values

artf (approved):
💜

PR #6555, opened 2025-07-01
Change dynamic style events behviour

mohamedsalem401:
Some changes to how style changes, particularly those tied to data sources, trigger events:

* **Enable Event Emission for Data-Bound Styles:** Allow events to be emitted when styles even if a style is bound to a data source. This will enable real-time UI updates as the underlying data changes.
* **Emit Events on Data Source Value Changes:** When a style's value changes due to an update in its bound data source (e.g., `em.Datasource.get('ds').setRecord()`), an event like `em.emit(change:style:STYLE_NAME)` should be emitted. This provides a clear signal that a specific style has been updated.
* **Fix Event Emission for Redundant Data Bindings:** An event should still be emitted even when binding a style to a data source with a value that is identical to its previous static value. For example, if `cmp.getStyle('width')` currently returns '100px' and you then set `cmp.setStyle({width: {type: 'data-variable', path: 'some_path.to.100px'}})` (assuming the data variable also resolves to '100px'), an event should still be triggered to indicate the change in binding type, even if the visual outcome is the same.
* **Fix Listening to Data Source Changes After Project Load:** Updating values had some bugs if the data source or record was added after the dynamic value was added to components.

### DIFF
--- a/packages/core/test/specs/data_sources/serialization.ts
+++ b/packages/core/test/specs/data_sources/serialization.ts
@@ -412,5 +412,109 @@ describe('DataSource Serialization', () => {
         },
       });
     });
+
+    test('should resolve styles, props, and attributes if the entire datasource is added after load', () => {
+      const styleVar = {
+        type: DataVariableType,
+        defaultValue: 'black',
+        path: 'new-unified-data.styleRecord.color',
+      };
+      const propAttrVar = {
+        type: DataVariableType,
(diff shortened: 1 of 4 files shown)


packages/core/test/specs/data_sources/serialization.ts
artf, line 1: 👏

artf (approved):

PR #6571, opened 2025-07-20
Fix parser to take account of `parser.optionsHtml.keepEmptyTextNodes` option

### LINKED ISSUE #6570: BUG: parser ignores `parser.optionsHtml.keepEmptyTextNodes` option
### GrapesJS version

Chrom v140.0.7307.0 (canary)

### Reproducible demo link

https://jsfiddle.net/y3gr69s4/

### Describe the bug

**How to reproduce the bug?**
1. Create editor instance with `parser.optionsHtml.keepEmptyTextNodes` option set `true`.
   ```js
     parser: {
       optionsHtml: {
         keepEmptyTextNodes: true,
       },
     },
   });
   ```
2. Load HTML code that contains newlines between elements.
   ```js
   editor.setComponents('<body>\n<p>foo</p>\n<p>bar</p>\n</body>');
   ```
3. Get HTML code.
   ```js
   const result = editor.getHtml();
   ```

The resulting HTML code still contains newlines between elements.

```html
<body>
<p>foo</p>
<p>bar</p>
</body>
```

Newlines between elements are stripped from the resulting HTML code.

```html
<body><p>foo</p><p>bar</p></body>
```

We get the expected result if we set `parser.keepEmptyTextNodes` option to `true` but it causes TypeScript type error.  I think that `keepEmptyTextNodes` should be a property of `optionsHtml` since it is irrelevant to CSS parser.

```typescript
  parser: {
    // Error: 'keepEmptyTextNodes' does not exist on type 'ParserConfig'
    keepEmptyTextNodes: true,
  },
});
```

### Code of Conduct

nanto:
According to type definition, `parser.optionsHtml` option has `keepEmptyTextNodes` property but setting the property has no effect.

This pull request makes HTML parser to take account of `parser.optionsHtml.keepEmptyTextNodes` option.

Fix #6570

### DIFF
--- a/packages/core/src/editor/config/config.ts
+++ b/packages/core/src/editor/config/config.ts
@@ -205,12 +205,6 @@ export interface EditorConfig {
   tagVarEnd?: string;
 
   /**
-   * When false, removes empty text nodes when parsed, unless they contain a space.
-   * @default false
-   */
-  keepEmptyTextNodes?: boolean;
-
-  /**
    * Return JS of components inside HTML from 'editor.getHtml()'.
    * @default true
    */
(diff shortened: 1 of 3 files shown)


artf (approved):
LGTM 👍 thanks @nanto

PR #6600, opened 2025-08-27
[#6587]: Fix infinite text nodes on dragging into a symbol

### LINKED ISSUE #6587: BUG: Adding text to a symbol replaces the text node over and over again
### GrapesJS version

Firefox Developer 142.0b9, Chrome 138.0.7204.184

### Reproducible demo link

https://jsfiddle.net/artur_arseniev/ta19s6go/

### Describe the bug

**How to reproduce the bug?**
1. Make the "Article" component a symbol
2. Add a text component to the original "Article" component
3. Directly try to edit the text or set the cursor inside the text

Cursor is not always resetting to position 0 and the text component is not flickering, behaving like a normal text component.

- The cursor is reset to position 0 and its almost impossible to select text
- Constant editor 'component:add' events are fired

After "leaving" the text component and re-selecting it, the problem is gone and no further component:add events are fired.
I think for some reason the newly created text component is constantly replaced it with itself.

I haven't had the time to look into the code yet, so currently no idea what could cause the issue.

If anything is missing or unclear, please let me know.
Have a nice day :)

### Code of Conduct

mohamedsalem401:
For #6587 

When adding a text node to a symbol, an **infinite loop** was triggered. This occurred because the `Editor.store()` method was being called, which in turn activated the text node's `sync` content method, causing a repetitive cycle of adding the text node and calling `Editor.store()` again.


artf (approved):
🚀

PR #6604, opened 2025-09-04
Escape ampersand, less-than, and greater-than in attribute value

### LINKED ISSUE #6603: BUG: Ampersand is not escaped in attribute value
### GrapesJS version

Chrome v142.0.7392.0 (canary)

### Reproducible demo link

https://jsfiddle.net/xce183nm/1/

### Describe the bug

**How to reproduce the bug?**
1. Create an editor instance and load HTML code that contains character references in attribute value.
    ```js
      components: '<body><p title="&lt; &amp;amp; &gt;"></p></body>',
    });
    ```
2. Get HTML code.
    ```js
    const result = editor.getHtml();
    ```

Symbols in attribute value are escaped with character references. The `title` attribute value of the `p` element is parsed as `< &amp; >`.

```html
<body><p title="&lt; &amp;amp; &gt;"></p></body>
```

Symbols in attribute value are not escaped. The `title` attribute value of the `p` element is parsed as `< & >`.

```html
<body><p title="< &amp; >"></p></body>
```

We have to escape ampersand to keep round-trip conversion between HTML parsing and serializing.

In addition, I think it is better to escape less-than and greater-than since:

1. Latest browsers escape less-than and greater-than in attribute value (e.g. `const p = document.createElement('p'); p.title = '< &amp; >'; p.outerHTML;` outputs `'<p title="&lt; &amp;amp; &gt;"></p>'`), and
2. There are many programs that process HTML code with [rough regexp pattern like `<[^>]+>`](https://github.com/search?q=%22%3C%5B%5E%3E%5D%2B%3E%22&type=code). Escaping less-than and greater than improves interoperability with those programs.

### Code of Conduct

nanto:
When serializing to HTML code, we should escape ampersand, less-than, and greater-than with character references to keep round-trip conversion and to improve interoperability with latest browsers and other programs.

Fix #6603

### DIFF
--- a/packages/core/src/dom_components/model/Component.ts
+++ b/packages/core/src/dom_components/model/Component.ts
@@ -11,7 +11,15 @@ import {
   bindAll,
   keys,
 } from 'underscore';
-import { shallowDiff, capitalize, isEmptyObj, isObject, toLowerCase } from '../../utils/mixins';
+import {
+  shallowDiff,
+  capitalize,
+  isEmptyObj,
+  isObject,
+  toLowerCase,
+  escapeAltQuoteAttrValue,
+  escapeAttrValue,
+} from '../../utils/mixins';
 import StyleableModel, {
   GetStyleOpts,
   StyleProps,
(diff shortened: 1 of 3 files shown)


artf (approved):

artf:
Thanks @nanto 🙇‍♂️

PR #6627, opened 2025-10-13
Fix: Prevent null reference error in drag-and-drop sorting

rainerregan:
Problem:
- Custom drag-and-drop operations were failing with DropLocationDeterminer.ts:254 Uncaught TypeError: Cannot read properties of undefined (reading 'clone')
- Occurred when component children existed in the model but didn't have valid DOM dimensions for sorting
Mismatch between [children.length > 0]
- Mismatch between `children.length > 0` and empty `childrenDimensions` array caused undefined `placeholderDimensions`

Solution:
- Adding condition to check nodeHasChildren and childrenDimensions.length > 0 before processing the positioning.
- Adding null check on nodeDimensions.clone() to make sure `placeholderDimensions` is defined.
- Prevent crashes when creating custom logic for drag and drop operation.


artf:
Hey thanks @rainerregan, looks like a legit fix but I'm wondering if by any chance you have an HTML snippet (`the model but didn't have valid DOM dimensions for sorting`) that could reproduce the issue

PR #6642, opened 2025-11-09
Support `attributes` option alongside `cleanId` for `toHTML` operations

DavidHarvey:
Currently the two options are mutually exclusive, this allows both to be applied.


artf (approved):
Thanks @DavidHarvey 🙌

Issue #6651, opened 2025-11-17
BUG: Droppable is ignored when moving a textable component into text


DavidHarvey:
### GrapesJS version

- [x] I confirm to use the latest version of GrapesJS

### What browser are you using?

Chrome 142.0.7444.61

### Reproducible demo link

https://codepen.io/DavidHarvey-the-lessful/pen/ZYWeMbG

### Describe the bug

Component `droppable` status is ignored when moving a `textable` component into a `text` component.

**How to reproduce the bug?**
1. Attempt to drag-and-drop a textable component into a text component with `droppable: false`.

**What is the expected behavior?**
As a user, I would expect that setting droppable to false would prevent dropping any components inside.

**What is the current behavior?**
Dropping is always allowed.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

artf:
https://github.com/GrapesJS/grapesjs/pull/6652#issuecomment-3551018995

PR #6652, opened 2025-11-17
Respect droppable status when moving a textable component into text

DavidHarvey:
Addresses https://github.com/GrapesJS/grapesjs/issues/6651

### DIFF
--- a/packages/core/src/dom_components/index.ts
+++ b/packages/core/src/dom_components/index.ts
@@ -917,13 +917,9 @@ export default class ComponentManager extends ItemManagerModule<DomComponentsCon
     if (isFunction(droppable)) {
       droppable = !!droppable(srcModel, target, index);
     } else {
-      if (droppable === false && target.isInstanceOf('text') && srcModel.get('textable')) {
-        droppable = true;
-      } else {
-        const el = srcModel.getEl();
-        droppable = isArray(droppable) ? droppable.join(',') : droppable;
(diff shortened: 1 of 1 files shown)


artf:
Hey @DavidHarvey, thanks for the PR, but I think this would break the textable functionality.
By default, we avoid dropping anything inside text components and it's false by default. The textable is intentionally breaks that rule on text components and that's why we have that logic here.

DavidHarvey:
@artf Do you have a suggestion for exposing the ability to prevent dropping textable components onto text in that case? If we're intentionally ignoring the droppable value of the destination, then should we have another configuration value?

artf:
@DavidHarvey I think, based on the current logic, you can extend the text component with droppable as a function:
```js
droppable: () => false,
```

DavidHarvey:
@artf That works as a workaround. However, are we ok with telling users that:
```js
droppable: false
```
has different functionality than
```js
droppable: () => false
```
in certain circumstances? This strikes me as a possible point of confusion.

artf:
Yeah agree, it's a bit confusing right now but textable prop is a bit "special" I'd say.

I'd avoid changing the default Text behavior but maybe we can introduce a new prop to Text component that allows to skip textable components inside (eg. `avoidTextable: false` as default), wdyt?

FYI I'll close the ticket and the PR at the moment

PR #6672, opened 2025-12-10
Fix incorrect conversion of rgba to hex8 in color picker

DavidHarvey:
Hex color codes with an alpha channel are formatted like `#rrggbbaa`, however the tinycolor implementation formats it as `#aarrggbb`. This fix places the alpha in the correct position at the end.


artf (approved):
Thanks @DavidHarvey 🙇‍♂️

PR #6681, opened 2026-01-06
Port newer color readability functions from TinyColor

DavidHarvey:
The "port" of TinyColor is either old, or modified, such that readability methods return unexpected results. This simply ports the latest readability methods (and necessary complementary methods, like `getLuminance`).

As an example, take the background color #0066cc and the foreground color #fff. This combination passes all WCAG contrast checks, except for AAA on normal sized text (See: https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=0066CC)

Before:

```js
-> tinycolor.isReadable('#0066cc', '#fff')
<- false
```

After:

```js
-> tinycolor.isReadable('#0066cc', '#fff')
<- true

-> tinycolor.isReadable('#0066cc', '#fff', {level: 'AAA', size: 'small'})
<- false
```

### DIFF
--- a/packages/core/src/utils/ColorPicker.ts
+++ b/packages/core/src/utils/ColorPicker.ts
@@ -1350,9 +1350,12 @@ export default function ($, undefined?: any) {
       RsRGB = rgb.r / 255;
       GsRGB = rgb.g / 255;
       BsRGB = rgb.b / 255;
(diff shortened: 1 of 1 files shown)


artf (approved):

artf:
Thanks @DavidHarvey 🙏

PR #6684, opened 2026-01-09
allow for more complex selectors in parseSelector

tomekczm:
this allowes you to have a selector with additional pseudoclasses like :not(.some-class-name) which would previously get interpreted as 2 seperate selectors

main benefit from this change is that if you add :not(.gjs-selected) to all your states in selectorManager, the default one becomes a preview of your component as stateless rather than a life preview, which also fixes a issue where if your component is in dragMode absolute and some state offsets the position, making the dragging glitchy


artf (approved):

artf:
Thank you @tomekczm

PR #6688, opened 2026-01-13
(Feat) custom attribute addition from the UI

krisbiradar:
# Feature: Custom Attribute Addition in Trait Manager

## 🚀 Overview
This feature enhances the Trait Manager by allowing users to dynamically add custom attributes to selected components directly through the editor UI. This removes the need to define every possible trait in the component definition, giving users the flexibility to add attributes like `data-*` tags, ARIA labels, or custom identifiers on the fly.

### 1. UI Updates
*   **New Input Field:** Added a footer section to the Trait Manager panel containing a text input and an "Add" button (`+`).
*   **Interaction:** Users can type an attribute name and add it by pressing `Enter` or clicking the button.

### 2. Core Logic
*   **Dynamic Addition:** When a trait is added, it is immediately registered on the selected component's `traits` collection.
*   **Attribute Initialization:** The feature automatically initializes the attribute on the component model (e.g., adding `data-id=""`), ensuring it appears in the exported code immediately.
*   **Validation:** Implemented a validation mechanism to prevent invalid attribute names.
    *   **Default Behavior:** Enforces `data-*` naming convention (e.g., `data-tooltip`).
    *   **Duplicate Check:** Prevents adding attributes that already exist on the component.

### 3. Configuration
*   **Custom Validation:** Introduced a new configuration option `customTraitValidationRegex` to override the default validation pattern.

## ⚙️ Configuration

You can customize the validation logic by passing a regex to the `traitManager` configuration.

**Default Behavior (Strict `data-*` attributes):**
```javascript
// No config needed, defaults to: /^data-[a-z][a-z0-9-]*$/

### DIFF
--- a/docs/api/datasources.md
+++ b/docs/api/datasources.md
@@ -113,6 +113,18 @@ const ds = dsm.get('my_data_source_id');
 
 Returns **[DataSource]** Data source.
 
+## getAll
+
+Return all data sources.
+
+### Examples
+
+```javascript
+const ds = dsm.getAll();
+```
+
+Returns **[Array][8]<[DataSource]>**&#x20;
+
 ## getValue
 
 Get value from data sources by path.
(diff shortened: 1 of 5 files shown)


artf:
Thanks @krisbiradar appreciate the contribution, but this could be done via a plugin, so there is no real need to add it to the core library.

PR #6753, opened 2026-04-22
Fix incorrect link to storage module

yogiduzit:
The components module contains an incorrect link to the storage module (storage instead of Storage).


artf (approved):

artf:
Thank you Yogesh

PR #6778, opened 2026-06-09
fix(asset-manager): enforce accept filter on drag-and-dropped files

### LINKED ISSUE #6032: BUG: Image Component allows upload and selection of video files via drag and drop
### GrapesJS version

Chrome v126

### Reproducible demo link

https://grapesjs.com/demo

### Describe the bug

**How to reproduce the bug?**
1. Open the [demo](https://grapesjs.com/demo)
2. Double-click on any existing image component to open the asset manager.
3. Drag and drop a video file into the image upload drop zone.
4. Wait for the video file to upload.
5. Click on the uploaded video asset to select it.

**Note:** This issue only occurs with the drag and drop. Video component also affected by this issue.

The Image component should prevent the selection of incorrect file types. It would be better if the asset manager blocks the uploading of incorrect file types as the first step.

### Code of Conduct

geonsang-jo:
Fixes #6032

## Problem
Dropping a non-image file (e.g. a video) onto an image component's asset upload zone uploads it and lets you select it as the image `src`. Reported as drag-drop-only.

## Root cause
`OpenAssets` sets `accept="image/*"` on the upload input, which filters the native file picker — but browsers do NOT enforce `accept` on drag-and-dropped files. `FileUploader.uploadFile()` reads `e.dataTransfer.files` and uploads them with no validation.

## Fix
Filter incoming files in `uploadFile()` against the input's `accept` value via a small pure helper (`isFileAccepted`), and bail out before uploading if none match. Backward-compatible: with no/`*/*` accept (generic asset managers), nothing is filtered. Custom `uploadFile` implementations are unaffected.

## Tests
Unit tests for `isFileAccepted` (MIME wildcard / exact / extension / `*/*`) and an integration test asserting a dropped video does not trigger an upload while a dropped image does.

## Live demo
Side-by-side StackBlitz (double-click the image to open the Asset Manager, then drop a video vs an image):
- Fixed: https://stackblitz.com/github/geonsang-jo/grapesjs-core-demos/tree/6032-image-video-accept?file=src/main.ts&startScript=dev&view=preview
- Before fix (bug repro): https://stackblitz.com/github/geonsang-jo/grapesjs-core-demos/tree/6032-as-is?file=src/main.ts&startScript=dev&view=preview

## Notes on behavior
- The rejection is silent, consistent with the existing `beforeUpload` -> `false` cancel path (which also returns without emitting an event) and with the native `accept` picker (which silently omits non-matching files).
- `beforeUpload(files)` now receives the filtered `File[]` instead of the raw `FileList` (type `(files: any) => void | false`, so type-compatible). Also, when every dropped file is rejected, `beforeUpload` is not invoked at all (the early return precedes it).

## Questions for maintainers
1. **Rejected-file feedback:** the AM has no "rejected/invalid" event today, and the existing block hook (`beforeUpload` -> false) is silent, so this PR keeps rejection silent. If you'd prefer consumers be able to react (e.g. show "only image files allowed"), I'm happy to add an opt-in event following the `asset:upload:*` convention, e.g. `asset:upload:rejected` with `{ files, accept }`.
2. **Stale `accept` on the singleton uploader:** `OpenAssets.run()` only sets `accept` when truthy and never resets it, and the uploader input is a singleton rendered once (`AssetManager.render()` builds `AssetsView` only on first open). So after an image picker sets `accept="image/*"`, a later generic `AssetManager.open()` with no `accept` inherits the stale `image/*` — before this PR that only affected the native picker, now it also makes drag-drop silently filter non-image files there. I can reset it on each open (`uploadEl.setAttribute('accept', accept || '*/*')`) here or as a follow-up.

### DIFF
--- a/packages/core/src/asset_manager/view/FileUploader.ts
+++ b/packages/core/src/asset_manager/view/FileUploader.ts
@@ -6,6 +6,26 @@ import html from '../../utils/html';
 import { AssetManagerConfig } from '../config/config';
 import { UploadFileClb, UploadFileOptions } from '../types';
 
+/**
+ * Check if a file matches an `accept` attribute value.
+ * https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#accept
+ * Allows everything when empty or a bare wildcard; otherwise supports MIME
+ * wildcards (`image/*`), exact MIME (`image/png`), and extensions (`.png`).
+ */
+export function isFileAccepted(file: File, accept?: string | null): boolean {
+  const acc = (accept || '').trim();
+  if (!acc || acc === '*' || acc === '*/*') return true;
+  const type = (file.type || '').toLowerCase();
+  const name = (file.name || '').toLowerCase();
+  return acc.split(',').some((raw) => {
+    const token = raw.trim().toLowerCase();
+    if (!token) return false;
+    if (token.startsWith('.')) return name.endsWith(token); // extension
+    if (token.endsWith('/*')) return type.startsWith(token.slice(0, -1)); // e.g. "image/"
+    return type === token; // exact MIME
+  });
+}
+
 type FileUploaderTemplateProps = {
   pfx: string;
   title: string;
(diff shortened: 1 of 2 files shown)


artf (approved):

artf:
Thank you @geonsang-jo Nice PR 👏

PR #6799, opened 2026-07-25
Deselect components on removal

### LINKED ISSUE #6696: BUG: Crash when calling component.replaceWith() during lifecycle events with UndoManager enabled
### GrapesJS version

Chrome

### Reproducible demo link

https://codepen.io/Logeshwaran-codepen/pen/raLJpYb

### Describe the bug

**How to reproduce the bug?**
1. Select components that available in canvas. Listen to a lifecycle event (e.g. component:selected)
2. Call component.replaceWith() inside the event handler
3. Perform undo

Undo should safely restore editor state without errors, or GrapesJS should provide a supported way to replace components during lifecycle events.

Undo crashes because GrapesJS tries to restore selection/state for a component that no longer exists.

**Question**
Is there a recommended, undo-safe way to replace a component during selection or status change events?

### Code of Conduct

Gxrvish:
Removing a component (directly or via replaceWith) left it in the editor's Selected collection, since only the core:component-delete command deselected explicitly. The stale entry breaks anything iterating the selection and, with UndoManager's trackSelection, lets undo restore a selection pointing at removed components.

Deselect in Components.removeChildren so every non-temporary removal path is covered. Temporary and undo-driven removals are unaffected.

Fixes: https://github.com/GrapesJS/grapesjs/issues/6696

### DIFF
--- a/packages/core/src/dom_components/model/Components.ts
+++ b/packages/core/src/dom_components/model/Components.ts
@@ -209,6 +209,7 @@ Component> {
       sels.remove(rulesRemoved.map((rule) => rule.getSelectors().at(0)));
 
       if (!removed.opt.temporary) {
+        em.removeSelected(removed);
         em.Commands.run('core:component-style-clear', { target: removed });
         removed.views.forEach((view) => {
           view.scriptContainer &&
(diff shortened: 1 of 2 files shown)


Gxrvish:
@artf Please have a look!!
Thanks!!

artf (approved):
well done @Gxrvish thank you also for the detailed test coverage 👏

PR #6804, opened 2026-07-31
fix(keymaps): apply prevent option without relying on CanvasView

### LINKED ISSUE #6695: BUG: Prevent default not working for edge
### GrapesJS version

Edge 144.0.3719.92 

### Reproducible demo link

https://jsfiddle.net/L61f0qtz/

### Describe the bug

**How to reproduce the bug?**

1. Open GrapesJS editor in Microsoft Edge.
2. Add a keymap for saving using either Ctrl+S or Ctrl+Shift+S with { prevent: true }.
3. Focus on the editor canvas (or anywhere in the editor).
4. Press the shortcut (Ctrl+S or Ctrl+Shift+S).

The GrapesJS command console.log("save") should execute.
The browser should not open the “Save” or “Save As…” dialog

The "console.log("save") not happen.
The browser triggers its default “Save” or “Save As…” dialog.
preventDefault() does not stop the browser shortcut in Edge.

### Code of Conduct

Gxrvish:
The `prevent` option delegated to `CanvasView.preventDefault`, resolved through `Canvas.getCanvasView()` with optional chaining, so the default browser action was silently kept whenever the canvas view was not available (editor not rendered yet, canvas removed). `Commands`, `Canvas` and the editor instance were also captured when the keymap was added instead of when it runs.

Move the prevention to a shared `preventDefault` DOM util, which also handles the original event of the canvas frame (`_parentEvent`), and resolve the modules inside the handler.

Fixes: https://github.com/GrapesJS/grapesjs/issues/6695

### DIFF
--- a/packages/core/src/canvas/view/CanvasView.ts
+++ b/packages/core/src/canvas/view/CanvasView.ts
@@ -12,6 +12,7 @@ import {
   isTextNode,
   off,
   on,
+  preventDefault,
 } from '../../utils/dom';
 import { getComponentView, getElement, getUiClass } from '../../utils/mixins';
 import Canvas from '../model/Canvas';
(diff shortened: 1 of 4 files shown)


artf (approved):
LGTM 👍

PR #6814, opened 2026-08-18
feat: allow component to have a lock ratio

UnderKoen:
Add an option to resize options that allows the component to keep it's ratio. 

This disables the shift behaviour and adds the ratio keeping to the center resize handles.

I was thinking maybe this should have multiple modes that would allow shift resize 

Or be split in 2 seprate options like `allowShiftRatioUnlock`, `enforceCenterRatio`

### DIFF
--- a/packages/core/src/commands/view/Resize.ts
+++ b/packages/core/src/commands/view/Resize.ts
@@ -253,8 +253,8 @@ export default class CommandResize extends CommandAbstract<
 
         const { store, selectedHandler, config, resizer, event } = updateOptions;
         const { keyHeight, keyWidth, autoHeight, autoWidth, unitWidth, unitHeight } = config;
-        const onlyHeight = ['tc', 'bc'].indexOf(selectedHandler!) >= 0;
-        const onlyWidth = ['cl', 'cr'].indexOf(selectedHandler!) >= 0;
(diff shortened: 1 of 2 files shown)


artf:
Make sense, but `ratioLock` feels too close to `ratioDefault` and ends up mixing a few separate behaviors into one boolean.

I wonder if it would make sense to add a new option as a function instead (eg. `keepRatio`) to give more flexibility
```js
resizable: {
  keepRatio: ({ handle, shiftKey }) => {
    if (['cl', 'cr', 'tc', 'bc'].includes(handle)) return true;
    return !shiftKey;
  },
}
```
wdyt?

UnderKoen:
Maybe a mix?

```ts
type keepRatio = 'ALWAYS' | 'NEVER' | 'DEFAULT' | (...) => boolean;
```

artf:
That makes more sense to me. My only doubt is DEFAULT: would it actually mean anything different from leaving the option unset/undefined?

UnderKoen:
Didnt really know how to name it, as it would also be weird to have no explicit option for the default behavior.

Maybe LOOSE

artf:
I'd prefer to have keepRatio simply undefined to indicate "default" behavior

Issue #6748, opened 2026-04-17
BUG: Dragging blocks crashes when template contains HTML comment nodes


hosseinghzadeh:
### GrapesJS version

- [x] I confirm to use the latest version of GrapesJS

### What browser are you using?

Chrome 147

### Reproducible demo link

https://jsfiddle.net/9rgxb3jk/

### Describe the bug

**How to reproduce the bug?**
1. Create a GrapesJS instance with a template that contains any HTML comment node (`<!-- ... -->`)
2. Add any block to the Block Manager
3. Drag the block into the canvas

**What is the expected behavior?**
Block drops into the canvas without errors.

**What is the current behavior?**
`TypeError: Illegal invocation` is thrown and the drag-and-drop crashes.

**Root cause**
PR #6692 removed `getDisplayedChildren()` which previously excluded non-Element nodes
via `element.nodeType === Node.ELEMENT_NODE` check inside `isDisplayed()`.
After this change, `getChildren()` returns all components including those backed by
comment nodes (`nodeType 8`). When the sorter calls `matches()` on these nodes during
drag-and-drop, it throws `TypeError: Illegal invocation` since `matches()` only exists
on Element nodes.

**Suggested fix**
Guard `matches()` calls in the sorter with a `nodeType` check to skip non-Element nodes:

```ts
if (node.nodeType === Node.ELEMENT_NODE && node.matches(selector)) { ... }
```

**Proof (run in JSFiddle console after loading the editor)**
```js
const iframe = document.querySelector('.gjs-frame');
const doc = iframe.contentDocument;
const walker = document.createTreeWalker(doc.body, NodeFilter.SHOW_COMMENT);
let node;
while (node = walker.nextNode()) {
  try { node.matches('div'); }
  catch(e) { console.error('matches() failed on comment node:', e.message); }
}
```
Output:
```
Comment node found: [if mso | IE]></td></tr></table><![endif]
matches() failed on comment node: node.matches is not a function
```

This simulates exactly what the sorter does during drag-and-drop, confirming that
`matches()` throws on comment nodes (`nodeType 8`). In our application, this manifests
as `TypeError: Illegal invocation` when dragging blocks into templates containing
HTML comments.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #6692, opened 2026-01-21
fix drag drop cause removing wrong element

them7d:
fix #6666 
This PR addresses issues with component sorting and movement by removing visibility filters in BaseComponentNode.
Previously, hidden components were excluded from the children list, which led to incorrect index calculations (cause removing element when invisible element in the same collection) when attempting to move or remove items (resorting the tree).

Additionally, CanvasNewComponentNode has been updated to correctly handle index offsets when checking or moving multiple components (e.g., arrays of content), ensuring that each item is processed at the correct position.

Changes:

- BaseComponentNode.ts: Removed getDisplayedChildren, isChildDisplayed, and the isDisplayed helper. getChildren now returns all components from the model, regardless of visibility.

- CanvasNewComponentNode.ts: Updated canMove logic to account for the index offset (index + i) when validating moves for array content.

PR #6818, opened 2026-08-19
fix(sorter): use model-space index when filtered children precede drop target

Amwam:
## Problem

When a container's model children include nodes that are filtered out of `childrenDimensions` (any node whose DOM element doesn't match the `itemSel` selector — most commonly `comment` nodes with `nodeType === 8`), the insert index returned by `getDropPosition` is off by the number of skipped children.

### Root cause

`getChildrenDim` iterates the full model-child collection but skips non-element nodes:

```ts
each(children, (sortableTreeNode, i) => {
  const el = sortableTreeNode.element;
  if (!isTextNode(el) && !matches(el, this.containerContext.itemSel)) {
    return; // comment nodes exit here — but i keeps incrementing
  }
  const dim = this.getDim(el);
  dims.push(dim);  // dims-array index ≠ model index i
});
```

`dims[0]` is model child 1, `dims[1]` is model child 3, etc. `findPosition` returns a `dims`-array index, but `getDropPosition` was using that directly as the model insert index:

```ts
({ index, placement } = findPosition(childrenDimensions, mouseX, mouseY));
index = index + (placement == 'after' ? 1 : 0);  // wrong — dims index, not model index
```

### Concrete example

`grapesjs-mjml` injects MSO conditional comment nodes (`<!--[if mso | IE]>...<![endif]-->`) as siblings of every `mj-section` inside `mj-body`. A body with 5 sections has 10 model children (5 comments + 5 sections). `childrenDimensions` has 5 entries (comments filtered). Dropping after Section 5:

- `findPosition` returns `dimensionIndex = 4`
- Old code: `index = 4 + 1 = 5` → block lands at Section 3's slot ❌
- Fixed code: `index = 9 + 1 = 10` → block lands after Section 5 ✅

**Live demos:**
- 🐛 Bug (grapesjs@0.23.5): https://jsfiddle.net/9aqgseyw/1/
- ✅ Fix (patched build): https://jsfiddle.net/1q7szjcr/

Drag "Drop me" after Section 5 in each — the log overlay shows the model index where the block landed.

## Fix

Two changes to `DropLocationDeterminer`:

**1. Stamp the model-space index onto each `Dimension` before filtering (`getChildrenDim`):**

```ts
const dim = this.getDim(el);
dim.dir = this.getDirection(el, targetElement);
dim.indexEl = i;  // preserve loop index (model-space) before filtering
dims.push(dim);
```

**2. Use `indexEl` instead of the `dims`-array index when computing the insert position (`getDropPosition`):**

```ts
const { index: dimensionIndex, placement: foundPlacement } = findPosition(childrenDimensions, mouseX, mouseY);
placement = foundPlacement;
const dimension = childrenDimensions[dimensionIndex];
placeholderDimensions = dimension.clone();
index = (dimension.indexEl ?? dimensionIndex) + (placement == 'after' ? 1 : 0);
```

The `?? dimensionIndex` fallback preserves existing behaviour when `indexEl` is not set (e.g. the legacy `legacyIndex` path).

## Prior art

- #6692 — same class of bug on the `getDisplayedChildren()` axis (filtered index applied to full collection)
- #6748 / #6750 — comment-node `nodeType` guard in `SorterUtils.ts` (fixes a `matches()` crash but does not fix the index skew)

### Problem example:
https://github.com/user-attachments/assets/31b07750-1ac4-4a9d-a223-b4093ac0d487

### Solution

https://github.com/user-attachments/assets/5444add3-3908-4a01-9b80-4f4df76777d2

### DIFF
--- a/packages/core/src/utils/sorter/Dimension.ts
+++ b/packages/core/src/utils/sorter/Dimension.ts
@@ -12,6 +12,8 @@ export default class Dimension {
   public width: number;
   public offsets: ReturnType<CanvasModule['getElementOffsets']>;
   public dir?: boolean;
+  /** Index of this child in the *unfiltered* children list (model-space). Set by getChildrenDim. */
+  public indexEl?: number;
 
   /**
    * Initializes the DimensionCalculator with the given initial dimensions.
@@ -34,6 +36,7 @@ export default class Dimension {
     this.width = initialDimensions.width;
     this.offsets = initialDimensions.offsets;
     this.dir = initialDimensions.dir;
+    this.indexEl = initialDimensions.indexEl;
   }
 
   /**
@@ -111,6 +114,7 @@ export default class Dimension {
       width: this.width,
       offsets: { ...this.offsets }, // Shallow copy of offsets
       dir: this.dir,
+      indexEl: this.indexEl,
     });
   }
 
--- a/packages/core/src/utils/sorter/DropLocationDeterminer.ts
+++ b/packages/core/src/utils/sorter/DropLocationDeterminer.ts
@@ -252,10 +252,14 @@ export class DropLocationDeterminer<T, NodeType extends SortableTreeNode<T>> ext
     let placeholderDimensions = nodeDimensions.clone(),
       index = 0,
       placement = 'inside' as Placement;
-    if (nodeHasChildren) {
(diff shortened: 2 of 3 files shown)


artf (approved):
Thanks @Amwam  for the fix and for the very detailed report 🙏

PR #6824, opened 2026-08-28
fix: drag&drop issues when the editor is in a shadowRoot

### LINKED ISSUE #6823: BUG: Drag&drop issues when the editor is in a shadowRoot
### GrapesJS version

Firefox 154 and Chrome 151

### Reproducible demo link

https://jsfiddle.net/dgLc7qz4/

### Describe the bug

**How to reproduce the bug?**
We are wrapping the editor in a custom webcomponent with a shadow root. See the reproducer

1. Drop some elements in the DOM
2. Try to re-order them in the layer manager

- A green placeholder appears when dragging an element in the editor
- Elements placed in the editor are sortable in the layer manager.

- The green "placeholder" doesn't appear - no visual indicator seems to appear at all
- The elements cannot be sorted in the layer manager

**Additional notes**
We had no issue in version 0.21.13, and it appear in version **0.22.1**. Our main suspect is this code block

```typescript
// src/utils/sorter/Sorter.ts (0.23.6, line 173)

private ensurePlaceholderElement() {
  const el = this.placeholder.el;
  const container = this.containerContext.container;
  if (!el.ownerDocument.contains(el)) {     // always false inside a shadow root
    container.append(this.placeholder.el);
  }
}
``` 

### Code of Conduct

fauvetg:
Basic implementation of the fix for https://github.com/GrapesJS/grapesjs/issues/6823

It contains two fixes for the case when the editor is loaded in a shadowRoot (in a webcomponent) : 
- Sorter#ensurePlaceholderElement > use `el.isConnected` instead of `el.ownerDocument.contains`
- Fix DropLocationDeterminer because `Document.elementFromPoint` does not pierce shadow boundaries

It compiles successfully and it resolves the issue I had. However I am not 100% that the fix is correct. Please let me know if you want me to improve it our change something

### DIFF
--- a/packages/core/src/utils/sorter/DropLocationDeterminer.ts
+++ b/packages/core/src/utils/sorter/DropLocationDeterminer.ts
@@ -4,7 +4,7 @@ import EditorModel from '../../editor/model/Editor';
 import { isTextNode, off, on } from '../dom';
 import { SortableTreeNode } from './SortableTreeNode';
 import { Placement, PositionOptions, DragDirection, SorterEventHandlers, CustomTarget, DragSource } from './types';
-import { bindAll, each } from 'underscore';
+import { bindAll, each, isFunction } from 'underscore';
 import { matches, findPosition, offset, isStyleInFlow } from './SorterUtils';
 import { RateLimiter } from './RateLimiter';
 import Dimension from './Dimension';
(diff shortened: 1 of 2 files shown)


artf (approved):

artf:
Thank you @fauvetg

